PR #6, opened 2019-04-04
Fix BookReader fullscreen

giancarlobi:
We have to override .BookReader.fullscreen to add !important.
Using @DiegoPino code, max width/height settings applyed to BookReader div


DiegoPino:
@giancarlobi wonderful :) Will review once my many Meetings end today. Looks good and this is your first official commit so 🏅 for you 👍

PR #10, opened 2019-04-09
First working External manifest

giancarlobi:
Maybe we need additional checks or code cleaning, while this works following these steps:
- add a display (i.e. Digital Object with Book Reader and external manifest) and set media source Strawberryfield JSON Key with a Manifest URL and Json key (i.e. iiifmanifest)
- add a digital object with a minimal Json with a key "iiifmanifest" with manifest URL value (i.e. http://www.byterfly.eu/islandora_iiif/librib:927474/manifest) and assign previous created display.
@DiegoPino feel free to wait for merging, probably I'll make more check tomorrow, as you want

### DIFF
--- a/js/iiif-iabookreader_strawberry.js
+++ b/js/iiif-iabookreader_strawberry.js
@@ -34,7 +34,7 @@
                         var options = {
                             ui: 'full', // embed, full (responsive)
                             el: '#' + element_id,
-                            iiifmanifesturl: 'https://iiif.archivelab.org/iiif/platowithenglish04platuoft/manifest.json',
+                            iiifmanifesturl: drupalSettings.format_strawberryfield.iabookreader[element_id]['manifesturl'],
                             iiifmanifest: drupalSettings.format_strawberryfield.iabookreader[element_id]['manifest'],
                             iiifdefaultsequence: null, //If null given will use the first sequence found.
                             maxWidth: 800,
--- a/src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
@@ -234,7 +234,7 @@ class StrawberryPagedFormatter extends FormatterBase implements ContainerFactory
       ],
       'manifesturl_source' => [
         '#type' => 'textfield',
-        '#title' => t('JSON Key from where to fetch a full IIIF manifest URL'),
+        '#title' => t('JSON Key from where to fetch an absolute full IIIF manifest URL'),
         '#default_value' => $this->getSetting('manifesturl_source'),
         '#states' => [
           'visible' => [
@@ -423,14 +423,13 @@ class StrawberryPagedFormatter extends FormatterBase implements ContainerFactory
       // A rendered Manifest
       $manifest = '';
       switch ($pagestrategy) {
-        case 'json_key':
-
-          break;
+	case 'json_key':
+		break;
         case 'manifesturl':
-          $elements[$delta] = $this->processElementforManifestURL($delta, $jsondata, $item);
-          break;
+		$elements[$delta] = $this->processElementforManifestURL($delta, $jsondata, $item);
+		break;
         case 'metadatadisplayentity':
-          $elements[$delta] = $this->processElementforMetadatadisplays($delta, $jsondata, $item);
+          	$elements[$delta] = $this->processElementforMetadatadisplays($delta, $jsondata, $item);
       }
 
 
@@ -462,10 +461,7 @@ class StrawberryPagedFormatter extends FormatterBase implements ContainerFactory
       */
 
       $elements[$delta]['#attached']['library'][] = 'format_strawberryfield/iiif_iabookreader_strawberry';
-
-
     }
-
     return $elements;
   }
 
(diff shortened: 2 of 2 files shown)


giancarlobi:
@DiegoPino I see here https://github.com/esmero/format_strawberryfield/blob/d9aeaa4897c6117d765a00891d7cfd82f7b43069/js/plugin.iiif-iabookreader_strawberry.js#L36 the priority is manifest then manifesturl. I think that is ok. When we send manifesturl we have to check no manifest array is passed. Do you agree?

js/iiif-iabookreader_strawberry.js
```diff
@@ -34,7 +34,7 @@
                         var options = {
                             ui: 'full', // embed, full (responsive)
                             el: '#' + element_id,
-                            iiifmanifesturl: 'https://iiif.archivelab.org/iiif/platowithenglish04platuoft/manifest.json',
+                            iiifmanifesturl: drupalSettings.format_strawberryfield.iabookreader[element_id]['manifesturl'],
```
DiegoPino: Hi, i would recommend testing if the value exists and is not empty, just in case. Just in case, we should really trust that our settings are well formed and to be honest if i ask you to do that i will need to do the same for the whole structure =)

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -705,6 +703,65 @@ public function processElementforMetadatadisplays(
...
+    $element = [];
+    $entity = NULL;
+    $nodeuuid = $item->getEntity()->uuid();
+    $max_width = $this->getSetting('max_width');
+    $max_height = $this->getSetting('max_height');
+
+    if ($this->getSetting('manifesturl_source')) {
+	      $manifest_url_key = $this->getSetting('manifesturl_source');
+      	if ($jsondata[$manifest_url_key]) {
+		      $manifest_url = $jsondata[$manifest_url_key];
```
DiegoPino: Question: do you think we need to allow many URLS in that key? Is there a need for that?

DiegoPino: Since we don't have a validation on the value, should we try to make sure its an URL? I like this function https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21UrlHelper.php/function/UrlHelper%3A%3AisValid/8.6.x to test that and the use statement is already there =) https://github.com/esmero/format_strawberryfield/blob/master/src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php#L18

js/iiif-iabookreader_strawberry.js
giancarlobi: I added URL check into StrawberryPagedFormatter.php, I think checks better before (in PHP) than here, do you agree @DiegoPino ?

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
giancarlobi: Answer: at this time I don't imagine need with more than 1 manifest per book.
Great function, applied to manifest URL check if is valid, code updated.

js/iiif-iabookreader_strawberry.js
DiegoPino: Yes. So let me understand. If formatter passes a manifest and there is no inline Manifest, URL will work right? If the full manifest is passed, URL is not used. Right? I know its my code but i need a double pair of eyes!

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
giancarlobi: I forgot I supposed an absolute URL for manifest so I modified variable description adding "absolute".

js/iiif-iabookreader_strawberry.js
giancarlobi: Yes, exactly as you wrote the code:
if (self.options.iiifmanifest != null) {
...
else {
// Ok, no full manifest, now try with the remote URL
...

DiegoPino: Perfect.. did i write that? jajja. Thanks for refreshing my mind.

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -423,16 +423,13 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
       // A rendered Manifest
       $manifest = '';
       switch ($pagestrategy) {
-        case 'json_key':
-
-          break;
+	case 'json_key':
+		break;
```
DiegoPino: Tiny tiny change (sorry) could you indent it with a only two spaces instead of 4 (2 spaces = 1 tab)? It is not a concern really but once i have to do DCS it will probably complain. I can merge anyway, but if you manage to do it before even better. Thanks so much!

giancarlobi: Sorry, I checked all indents are 2 spaces, I hope it's ok now.

PR #15, opened 2019-07-20
Expose metadata to download and view directly

DiegoPino:
# What is new

Adds a new config entity to expose Twig template output directly as a URL on each node.

This is something i really wanted for long and will be happening for this release.

 Now that our Metadata Display Entities have a configurable format (application/json, text/html, etc) we can expose direct access URLs  on selected Content Types using a configurable and unique URLs.
Something like /do/uuuid/metadata/iiif/ds.json, where iiif is the machine name of the configuration entity and ds a given filename and .json the corresponding extension computed from metadata display entity's used to process the json into other formats. Hu, long phrase.

I still need to work more on this before monday and remove code and test. bBut looking good already, its the basis with some stub code yet and missing response controller. Brain dead so going to sleep, also will be 42 years old on Sunday and want to be alive for that!

`@TODO:` Bring the SBF Twig processing out of the formatter and into a ::service so we can reuse it here and in our new programatic views /rest endpoints.


giancarlobi:
@DiegoPino this is fantastic!! Especially regarding interaction with other repo, harvester, metadata collector, to provide oaipmh end point, iiif end point and much more.

giancarlobi:
@DiegoPino happy birthday!!!!

PR #16, opened 2019-07-26
Adds new Formatters, endpoints and download routes: Cleaning the field for 8.x-1.0-beta1

DiegoPino:
# What is new

So much!
This includes:
 - 3D Formatter for *any* type
- Added thumbnails to Open Seadragon
- Video
- Audio
- New Twig template configuration Entity to expose routes that render directly metada
- Bump to Drupal 8.7
- Direct route to download files not managed by Cantaloupe or any other IIIF image server, but keeping a IIIF aware path
- OCF helper should have never lived here, moved over the border to SBF land

@giancarlobi this brings the last piece into 8.x-1.0-beta1. I have code to clean and to update yet before the actual release, but this is a big advance


giancarlobi:
@DiegoPino wow!! what a wonderful work.
This and other PR mean that we have to update docker with all strawberryfield (STB, format, webform) branch 8.x-1.0-beta1, right?

DiegoPino:
Hi! thanks! I think docker should be named 1.0.0-beta1 wihtout the 8.x, since there is also solr, etc =) How do you feel? Or should be stick with same versions for all?

Cheers!

Diego Pino Navarro
Metropolitan New York Library Council
599 11th Av. New York, NY 10036

giancarlobi:
@DiegoPino I think that it's better the same name for all to avoid confusion, this is my external idea I don't know if there is something that doesn't allow this.

DiegoPino:
Yep. Makes sense! So lets keep docker in 8.x-1.0-beta1 and mark minio vesion v/s the full one somehow. Thx. Can you check if the pushed packages to packagist are there?
This was not working 20 minutes ago

docker exec -ti esmero-web bash -c "composer clear-cache; composer require strawberryfield/strawberryfield:dev-8.x-1.0-beta1 strawberryfield/format_strawberryfield:dev-8.x-1.0-beta1 strawberryfield/webform_strawberryfield:dev-8.x-1.0-beta1"  but maybe its just slow moving? (or i typed something wrong??

I just updated a running (AWS) docker instance from 8.6.12 to 8.7.5 and everything new and it worked perfectly!

giancarlobi:
@DiegoPino here https://github.com/esmero/webform_strawberryfield/blob/5a0f8f9c5fd157fbeaaa58d1a95df9419e2ff8d6/composer.json#L18 a typo error, beta instead of beta1

PR #17, opened 2019-07-26
Iab zoom

giancarlobi:


js/iiif-iabookreader_strawberry.js
```diff
@@ -40,10 +40,138 @@
...
                         var br = new BookReader(options);
                         br.init();
                     }
 
                 })}}
 })(jQuery, Drupal, drupalSettings);
+
+// override setupTooltips()
+// to enable system tooltip
+BookReader.prototype.setupTooltips = function() {
```
DiegoPino: This is awesome. I will merge this to have this awesome functionality in our book reader. Next step could be to move all Prototype functions maybe into https://github.com/esmero/format_strawberryfield/blob/master/js/plugin.iiif-iabookreader_strawberry.js, what do you think @giancarlobi ? Would make the main drupal js cleaner?

DiegoPino:
@giancarlobi this works great. I will add a Pull description to this so we have the record and will merge. Once i do a PHPCS tomorrow on the whole code base i will also try to move all bookreader.prototype functions into plugin.iiif js. Could be also a new plugin file if you feel that would be cleaner. This is awesome!!

Issue #20, opened 2019-10-03
Make a Panorama Tour Builder/+ Linked Digital Objects Aware Formatter


DiegoPino:
# Use case?

Building Tours using Panoramas, adding hotspots and bind them together is every Developers dream..eh nightmare..well, still a dream.. But, we can do this.

# How

Deal with two use cases. Single image Panoramas, and Multi Image Panoramas with Hotspots all over the place that allow you to run in circles. But, say no more, lets make this even more LoD aware, and allow Tours to be build using Single Image Panoramas based Digital Objects (our ADOs).

# When? Who?

I did it. But i'm too tired to explain the how. In the meantime i will push some Code and lets see tomorrow.

DiegoPino:
Resolved via https://github.com/esmero/format_strawberryfield/commit/4fb29c995dcccc922ac6a18214acbee8904e4678

PR #21, opened 2019-10-03
ISSUE-20: Tour/Scene builder Multi panorama aware Formatter.

DiegoPino:
# What is new?
see #20 
- New Pannellum JS version (2.5.4) only 22 days old!
- Tested Dynamically loading Libraries via Ajax via Ajax again. That was failing (of course, who does that) but know it, now it works. Short answer, Document context of JS does not exist when there is an in memory document. Who would have guessed.
- StrawberryPannellumFormatter now gets another setting. With that it can try to fetch a nested list of other Scenes and their hotspots, where each scene is a node id. Basically a Formatter that uses other Nodes and their Formatters. It gets render arrays from each Node using the same setting of the one you/me/someone is viewing (so smart...gosh...i wish i was me) and rebuilds a new render `array` with all the results. Pushed that data into a little bit modifed JS and BUM! You have a multi scene, multi hotspot panorama!

~~This is not ready and not beautiful but enough to discuss over some coffee. And well, it does work~~

I will mark this as good enough for our needs. If any other issues appear after this we can deal with them in a further pull

### DIFF
--- a/config/schema/format_strawberryfield.schema.yml
+++ b/config/schema/format_strawberryfield.schema.yml
@@ -94,6 +94,8 @@ format_strawberryfield.formatter.strawberry_pannellum_formatter:
       type: string
     rotation:
       type: string
+    autoLoad:
+      type: boolean
 format_strawberryfield.formatter.settings.strawberry_video_formatter:
   type: config_object
   label: 'Specific Config for strawberry_video_formatter'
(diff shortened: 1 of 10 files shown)


DiegoPino:
I rebased this one to 8.x-1.0beta1. Should be cleaner now. More pulls comming since i'm not totally convinced about the tour JSON structure yet. Works but data be improved/made simpler.

PR #29, opened 2019-10-23
ISSUE-18b: Fix Broken IIIF Formatters after ISSUE-18 refactor

DiegoPino:
# What was broken?

https://github.com/esmero/format_strawberryfield/pull/19 introduced some bugs, mostly stuff that  was working already. Incorrect calls helper methods, DCS issues all over the place and also some weird logic. I also addressed some things that were working (good) but not totally correct (could be done better)

## In specific?
*Restores IIIF General setting form to working condition*
This changes also assumptions because we need this to be readable:
if the method is named checkURL then we assume return TRUE means its valid, so the var that gets the result should not be say error, but valid. Also, the "type" was a random string (trying to match the formatter settings form element) that was not documented. Breaking code like this is easy. So we have now 2 class level constants.

*Restore IiifUrlValidator to working condition*
Cleanup DCS, document functions, add two constants, make checking the URL  less of an act of faith (check if who responds is a Cantaloupe server, only avoid checking localhost IIF external URLS if Drupal is also is running in localhost mode . Give messages some context and make sure ::checkUrl only responds with TRUE of FALSE. (there is no true of false. We use TRUE of FALSE ). This passes phpcs --standard=Drupal

*Update StrawberryBaseFormatter.php*

Give each method proper `PHPdocs`.
Remove unused USE statements, fix wrongly concatenated t() statements and use the correct placeholder replacements.
Change all false and true to FALSE and TRUE and fix the Set Error By name to make sure we are really showing the error inline and not somewhere invisible on the top of the page when setting things. Use the Constants when calling validator and don't use/assume element name as a valid argument, we have now valid class Constants. Lastly, rename an unused (for now) method to correct camelcase (that was left from old code, not introduced here)

*Fix broken formatter, regression*
Make sure that getImageSizes is being called with the iiif identifier (this is a regression error!)
Add the missing `/` in the iiif server img path since after first fix was still broken.
Note: Static image was not broken, i just panic-ed

# Interested Parties
@marlo-longley since this broke via your pull. My bad to not double/double check but we need to work on code quality. I feel smaller pulls will be our strategy until you get the handle of doing 

Test this code, using all Formatters please and make sure logic for IIIF thing is working. We can't fail again and i fear i need to move to other code now.

`phpcs` and also reading Drupal core code to see how you are doing things.
@giancarlobi since your sharp eyes detected the error i introduced by merging #19 
Any help you can give me here (your expert eyes).

We need to merge this ASAP people!  Thanks


DiegoPino:
@marlo-longley @giancarlobi as i explained in my last commit, we need to decide if servers end or not with a / at the end, and that consistency applies to saved settings and fetched settings. Seems to me that Twig template building is easier (and we depend on that right now) when / is at the end.
Thanks. I have to go home now, hope we can get this merged soon. If not i will revert ISSUE-18 until we have a full solution.

giancarlobi:
> @marlo-longley @giancarlobi as i explained in my last commit, we need to decide if servers end or not with a / at the end, and that consistency applies to saved settings and fetched settings. Seems to me that Twig template building is easier (and we depend on that right now) when / is at the end.

@DiegoPino @marlo-longley I vote for easier things so ok for / at the end of server url. Could be for a nice view omitted when rendered? Do this make sense?

DiegoPino:
@giancarlobi @marlo-longley i think its all about usability and consistency and in specific here, i will vote for having a stored/retrieved version with slashes always. 
One of the main reasons for this is that it is a base URL. Means it servers no purpose other than to have some other stuff appended at the end.

Means i will add something that deals with that, i kinda feel i have the code already somewhere (islandora matomo module i think) 
I agree also we could leave, for cosmetic purposes the non / version, but it would fear that could lead to confusion for people

In any case, i would vote for merging this (if no other problems detected by any of you both) and then doing the slash  normalization s a second pull. We need to have archipelago beta1 back to working.

Maybe also time to start documenting this decisions: question is where?

giancarlobi:
@DiegoPino I fully agree with stored/retrieved version with slashes always also when rendered to avoid confusion and due to it's a BASE url.
For me ok to merge to solve this breaking issue with current beta release asap.
The decision is related to this variable so could be in code or/and into the description of the text box, I think there will be sufficient, imho.

DiegoPino:
Ok, i think there was enough time. Will merge. Thanks to you all for testing. If there are more issues we deal with them on the next iteration.

Issue #22, opened 2019-10-09
Implement Actual Rendering of Exposed Metadata Endpoints


DiegoPino:
# What?

We already created on beta1 our exposed and configurable Metadata endpoints reusing the Metadata Display Entities (Twig). But the actual rendering was still missing. 

# What is needed?

A Content-type response aware Controller, that validates output before responding and deals with Twig template rendering

#What will be also included?

We will also inject into the Twig rendering context configurable Entity Reference Views (and with that also our Custom/Super Cool Solr ones!) and the IIIF endpoints as variables.

This will allow us to make more complex templates that benefit of pre-processed and aggregated data coming from other sources. Specific needs like Collection Manifest in IIIF for example, where we can pass as an argument, using the current Node, the data output of a `member` nodes list.

Suggestions are open to what other data structures, pre-computed, could be inject. The views part will be part of each ExposedMetadata Configuration entity configuration.

DiegoPino:
Almost done via #27 #31 #32 (x3!)
Still need the View Injection. Will work on that soon

DiegoPino:
Resolved and implemented

PR #32, opened 2019-10-31
ISSUE-22: Fix Routing issues and Active/inactive, Gives the Config Entity some Love

DiegoPino:
# What is new?

Bug fixes and enhancements
improvements on exposed endpoints

See #27 and #22 and #31 

- This adds a better Collection Listing format for each Config Entity, 
  - We also now (cool) show a demo/example link on the listing, so people can see/where/how each endpoint can be found. I like it!
- Makes sure `isActive()` is enforced in config but also on the Controller
- Adds also the Config Entity as a Cacheable dependency so changes like enable/disabling are visible immediately.
-  removes some wrong routing (no canonical for the config entity!). 

@giancarlobi this is all yours, please test if you can before merging. I feel i finally addressed everything. @marlo-longley if you want to look at this too!

### DIFF
--- a/format_strawberryfield.routing.yml
+++ b/format_strawberryfield.routing.yml
@@ -63,7 +63,8 @@ format_strawberryfield.iiifbinary:
     uuid: "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
     format: .+
     _entity_access: 'node.view'
-# Direct access to Metadata display processed json
+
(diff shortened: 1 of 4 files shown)


marlo-longley:
@DiegoPino Thanks for this new functionality. I just tested using this pull and also strawberryfield module branch [here](https://github.com/esmero/strawberryfield/pull/48) (thanks for your help with that). This is awesome and works great from what I can see!!

DiegoPino:
Thanks folks!

Issue #33, opened 2019-11-06
Trigger View Mode Selection based on ADO type for canonical Node URIs


DiegoPino:
# What is the problem?

Right now we allow users to select, manually, which of the many View Modes they want to use on an ADO (Archipelago Digital Object)/Node display page (the canonical URL). This has some benefits because it allows people to choose "wisely" and customize per ADO which on will be used. Some use cases also support that, since some schema.org types like "Book" could eventually be shown as a mix of audio and a PDF? and not only via the View Mode that provides the IA Book reader. But even so, it speaks against other uses cases, simpler ones, like a User Role that has / does not need to make that decision and does not even Know what each View Mode entails.

# The Solution

Build a simple config option that allows a Mapping between existing (Solr indexed) JSON Types values and defined in Drupal View Modes. This mappings will only kick in if during ingest/edit a Node was published using the Default View Mode (which implies also the default behaviours). And that also means, a power user can always override these mappings.

DiegoPino:
Resolved via https://github.com/esmero/format_strawberryfield/commit/4bf717c9a10a5cc624a13d8fee4356cefd76f713 Thanks @giancarlobi for testing!

DiegoPino:
I was aware of some UI inconsistency on having the add button on the bottom of the form on not grouped with the ADO and View Mode selection. I got feedback on Friday on this so decide to fix it. Reopening now.

DiegoPino:
Solved as fixed

PR #34, opened 2019-11-06
ISSUE-33: JSON Type/ADO  to View Mode Mapping

DiegoPino:
# What is new?

See #33 

This pull implements a generalized solution for triggering existing Drupal View Modes based on mapped settings between those and **existing** JSON types values indexed already in Solr.
High emphasis on:  **existing** JSON types values indexed already in Solr.

# How?

The logic is quite simple:
1. An Archipelago Settings Form Endpoint where you can select a pair of Types (select box populated from already ingested types) and a View Mode.
  - The form allows to add more pairs without limiting (for now) having the same repeated (@TODO) a few times
 - People can reorder how those mappings are evaluated. Useful for defining general types like "image" as defaults when not other more specific TYPE matches. Note: Image, e.g comes from a deeper JSON structure, so really anything that has a file of type image will match that one! V/S Photograph, e.g, that will be the Media Type field selected by a user during ingest using a Webform.
- People can delete individual mappings
- People can deactivate, without deleting, mappings
- People of course can Save the settings

2.-  A Hook, in specific `hook_entity_view_mode_alter()` that checks if the node has no custom View Mode setup and the URL that triggered is the canonical (in our case /do/uuid but will also work for /node/somenumber) and if the Object contains one or more SBF  (Strawberryfield) fields and has a "type" field property in any of those Fields (future, when people, e.g have more SBFs for other type of metadata, who knows!). If nothing matches it will just use the Default, if something matches after checking if there is a config with a pair that matches, and the order of evaluation is right, it will change the View Mode on the fly.

# What is missing/can be improved here?

There are some TODOs that i will tackle on a second pull and don't affect function right now:
A. We are only allowing existing Types to be added. Maybe we want people to add Types are not even present yet? That can be done manually really via the form, as we expose textfields, but maybe we want to have another source of TYPES somewhere.
B. the Table drag is showing textfields and in particular for the View Mode, the Machine Name. That is also Ok, for now, but it would be prettier to change that to rendered User facing names with a hidden form element so submit/add/delete still works. Very reason i went for know for a full textfield
C. We have no validation. Not because its complex, but not sure if i want to validate here, since we could want to allow people to import settings or even have things setup upfront before the view modes or even the Types are present.
D. Triggering happens only on the canonical. Maybe we want the same on Revisions but mostly on the "Preview" while ingesting/editing an ADO too.
E. We could also want to "hint" powerusers that will have access to set manually a View Mode and enforce that, which one is the recommended one, by setting in the exposed DS View Mode Switch/setup field on the Entity Form any of our mappings as "preselected" or default. That way even Admins could benefit of this, specially if on the run and forget to set the right view mode, but still override if needed. Since during Ingest, type is not set yet until they press "Save Metadata", the change should probably happen there.

# What is next?

We need to ship a default mapping /configuration file with our archipelago-deployment

# Screenshots

Settings page:

settings page

Export Config/sync page
export settings page

# How to test?

Create any Digital Object, select some Media Type from the drop list, lets say a 3D Model. fill the form, add an Image and a .STL file, select "Default" from the Display Settings Fieldset 
View-mode-selection
Set State to Published,
Save Metadata and then Save the Node. 

You should see an Image Viewer Instead of a 3D Viewer, since our current Default is setup to use the Open Seadragon Plugin.

Now fetch this branch. Clear Caches. Go to http://localhost:8001/admin/config/archipelago/viewmode_mapping
Add a new mapping between
ADO Type: 3DModel
and
View Mode: Digital Object with 3D Viewer. Press the "add button" at the bottom
Press Save button at the bottom

Now navigate to the Node/ADO just ingested. You should see now the 3D Viewer instead of default Openseadragon one.

@giancarlobi @marlo-longley @mitchellkeaney (last human tagged here: no need to act on this, just feel free to comment if what this does and how it happens is unclear)

### DIFF
--- a/config/schema/format_strawberryfield.schema.yml
+++ b/config/schema/format_strawberryfield.schema.yml
@@ -106,3 +106,33 @@ format_strawberryfield.formatter.settings.strawberry_video_formatter:
       type: integer
     posterframe:
       type: string
+# Multiple JSON type / View Mode Mappings
+format_strawberryfield.viewmodemapping_settings:
+  type: config_object
+  label: 'Archipelago View Mode to JSON Type Key mapping Configurations'
+  mapping:
+    type_to_viewmode:
+      type: sequence
+      label: 'JSON type key to View Mode Mapping'
+      sequence:
+        type: format_strawberryfield.viewmodemapping_settings.mapping
+
+format_strawberryfield.viewmodemapping_settings.mapping:
+  type: mapping
+  label: 'JSON Type Key/ View Mode pair'
+  mapping:
+    jsontype:
+      type: string
+      label: 'JSON Type Key'
+    view_mode:
+      type: string
+      label: 'View mode'
+    active:
+      type: boolean
+      label: 'Whether this Config pair is enabled or not'
+    weight:
+      type: integer
+      label: 'Order in which this is evaluated'
+
+
+
--- a/format_strawberryfield.links.menu.yml
+++ b/format_strawberryfield.links.menu.yml
@@ -17,6 +17,13 @@ format_strawberryfield.iiif_admin_settings_form:
   description: 'Configure IIIF Server'
   parent: strawberryfield.group.admin
   weight: 99
+#Adds to Archipelago a global IIIF admin settings
+format_strawberryfield.view_mode_mapping_settings_form:
+  title: 'ADO type to View Mode Mapping Form'
+  route_name: format_strawberryfield.view_mode_mapping_settings_form
+  description: 'Configure how ADO type defined in your SBF JSON can be magically mapped to Drupal Content View Modes.'
+  parent: strawberryfield.group.admin
+  weight: 100
 #Adds to Archipelago Expose Metadata Config
 entity.metadataexpose_entity.collection:
   title: 'Configure Exposed Metadata Endpoints'
--- a/format_strawberryfield.module
+++ b/format_strawberryfield.module
@@ -3,7 +3,8 @@
  * @file
  * Contains formater_strawberryfield.module.
  */
-
+use \Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\Core\Entity\EntityInterface;
 /**
  * Implements hook_page_attachments().
  *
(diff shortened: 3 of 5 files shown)


DiegoPino:
@giancarlobi thank you for testing! That extra '!' in type field could be coming from some test ingest. Since that SELECT box is using Solr, you can facet agains that Field in Solr and find the culprit, maybe its some error from us. There is still work to be done with this one, but i will prefer to have a second pull do that:
1.- Add a config form to define which Solr field will be the one providing the "type" values
- This probably means also having a top key only TYPE solr field, instead of every Type as we have now!
2.- Give admins a visual "hint" so they know why objects look like a certain View Mode, if the selected value is "default". That is really optional but i guess its not bad


Thanks again!

giancarlobi:
> @giancarlobi thank you for testing! That extra '!' in type field could be coming from some test ingest. Since that SELECT box is using Solr, you can facet agains that Field in Solr and find the culprit, maybe its some error from us.

@DiegoPino I checked solr but I found only this:
image
So the "!" is not indexed. I'll check it deeper.

DiegoPino:
@giancarlobi i hardcoded  solr field 'type_1' for this commit as the source for the select (shame on me! but it was needed as a first step), that really needs a config form since by default docker ships with that but not every installation needs/wants/has to use that field. Maybe you have two 'type' solr fields? I will check the code too, maybe there is a bug, i'm known for introducing bugs ajjajaa

DiegoPino:
Here is the line
https://github.com/esmero/format_strawberryfield/pull/34/files#diff-404d95dfe97da5dc8d2ce72a8f908d62R179

giancarlobi:
@DiegoPino I know you hardcoded "type_1" and I fully agree with a config param for this so don't spend your time about this, we'll check this later.
From solr query I have this fields including type_1 (from different objects):
```
"ss_type_1":"digital_object",
"sort_X3b_en_type_1":"0:6:P*@\u0005\nF,<2.P\u0000",
"sort_X3b_und_type_1":"H8FPF6L*H8\u0000",
"sort_X3b_en_type_1":"H8FPF6L*H8\u0000",
"sm_type_1":["Create",
          "Image",
          "Photograph",
          "Service"],
```
Maybe this could be useful for future debug.

PR #35, opened 2019-11-12
ISSUE-33: Small fix for empty Viewmode mapping configs

DiegoPino:
# What is new?

When we have no mappings yet, the alter hook throws some errors. Make sure mappings, even when empty are always an array() . Thanks @mitchellkeaney for testing and catching this

# How to test?

With existing mapping all should work as before, with an empty config, no error should appear.


DiegoPino:
@mitchellkeaney if you want to review feel free, if you trust me, also ok!

DiegoPino:
@mitchellkeaney and now (backchannel) you know how to test! 🏆

PR #36, opened 2019-11-24
ISSUE-33: make UI consistent

DiegoPino:
# What is new?

See #33 for context

After getting more feedback during a call on Friday i decided to address the fact that our "add" new mapping button for the ADO type to View Mode Form was on the bottom (unexpected) and not just right by the select boxes. This pull fixes that. It also changes the way we display the already selected View Modes and deals (sort of) with the fact that people could have deleted a view mode after we saved the mappings. If that happens, on load we replace the View Mode Label colum with a select box allow the Admin to fix its mistake. I also did some code rearranging (to avoid long lines, nothing really impressive there but adds some clutter to the pull, sorry!)

# How to test

Fetch branches, switch to this, clear caches, check your form.

If you want to test the behavior on a broken View Mode, you can either import a wrong Config (Note. i had to re-ident the config file after exporting, not sure why... but i had to. It did not change the syntax but maybe i made an error on the schema declaration? Weirdly all internal checks and flags/things pass) or delete a view mode and check. You should get a Select Box to select a new view mode and a mostly harmless message, that maybe could need an alert CSS class or something to make it more "dangerous" to the UI aficionado.

@giancarlobi @marlo-longley

### DIFF
--- a/src/Form/ViewModeMappingSettingsForm.php
+++ b/src/Form/ViewModeMappingSettingsForm.php
@@ -56,6 +56,9 @@ class ViewModeMappingSettingsForm extends ConfigFormBase {
     $form['info'] = [
       '#markup' => $this->t('This Form allows you to map ADO (Archipelago Digital Object) "types" to existing Drupal View Mode Configurations.'),
     ];
+
+    $view_mode_list = $this->getViewModes();
+
     $form['add_fieldset'] = [
       '#title' => $this->t('Add a new mapping'),
       '#type' => 'fieldset',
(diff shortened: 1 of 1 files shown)


DiegoPino:
Here an example config YAML file to test non existing View Mode behavior
```YAML
--- 
type_to_viewmode: 
  - 
    active: true
    jsontype: Photograph
    view_mode: Default
    weight: -10
  - 
    active: true
    jsontype: Panorama
    view_mode: digital_object_with_pannellum_panorama_
    weight: -9
  - 
    active: true
    jsontype: 3DModel
    view_mode: digital_object_with_3d_viewer
    weight: -8
  - 
    active: true
    jsontype: Book
    view_mode: digital_object_with_book_reader
    weight: -7
  - 
    active: true
    jsontype: Document
    view_mode: collection_listing
    weight: -6
  - 
    active: true
    jsontype: Future Type
    view_mode: some_non_existing
    weight: -5
````

DiegoPino:
Hi Folks. Will move forward with this. I want to provide a default config for beta02 and this is needed. If something ends being wrong, we can fix as always.

Issue #38, opened 2020-01-22
Enable endpoint to serve temporary files to its owner


DiegoPino:
# What is needed?

During a webform ingest process (or any other that involves uploading a file to temp:// storage) files living in S3:// can not be served directly if Minio.io or S3 itself is inside a docker container using a docker local domain or if you are working on a development environment (localhost).

So we need a way of serving a resource, hopefully via a IIIF (fake) Drupal URI until it gets totally preserved and saved, and then we can just ask our IIIF server for the thing.

This could really go in any other place, like directly into SBF, but for consistency, and since we already have a `format_strawberryfield.directiiifbinary` route, makes more sense here.

So why not directly that existing one? that one depends on having an actual ADO published, it checks access to the node and if its an SBF bearing one. For a `temp` that is not even possible

DiegoPino:
Solved via https://github.com/esmero/format_strawberryfield/commit/6e2eca6e1183ba7f30327d1660b8a5fa8403606e

Issue #37, opened 2020-01-21
Regression bug: Formatters Config Schema is wrong


DiegoPino:
# What is happening?

This is a regression bug that slipped under my nose. After the unified IIIF settings and general IIIF form (base formatter) settings where introduced and merged, and a new schema was committed.  I just noticed that the schema was/is wrong and fails.

# How?

The schema for formatters, and in specific the one for specific ones defined here
https://github.com/esmero/format_strawberryfield/blob/master/config/schema/format_strawberryfield.schema.yml#L27 references %parent.type as source for the more specific configurations, which is wrong. Why?
First, there is not a single `type` key present in the `mapping` key inside `format_strawberryfield.formatter.*` . The example is quite clear here 
https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&ved=2ahUKEwiUsb-27pPnAhXOVt8KHTjlAKMQFjAAegQIAhAB&url=https%3A%2F%2Fwww.drupal.org%2Ffiles%2FConfigSchemaCheatSheet1.2.pdf&usg=AOvVaw264R1CNGzJke1Dbpn_hIaF

So we can not deduce which config should be used, but more over, the way the settings are defined are the other way around, means the specific ones override the general one, not extend it. More over, there seems no call/getter/setter for `formatter_config` key  in the forms we provide, leading to a completely wrong config. 

This is a regression bug since we had this working. 

I feel the solution here is more than simple but adds some extra lines. Let's forget the more general 
format_strawberryfield.formatter.* and simply move the properties that are inside the mapping to each individual, specific formatter. For now that will suffice. Also, kinda makes sense since not all Formatters require IIIF and right now the dependencies are in the wrong order.

This was introduced in https://github.com/esmero/format_strawberryfield/pull/19/files. 

I will make this as a blocker for the release but the fix is simple so should be done tomorrow

@marlo-longley @giancarlobi

DiegoPino:
Resolved by https://github.com/esmero/format_strawberryfield/pull/39/commits/ec4666868131d4b01fc94331051140f932147b8f

PR #39, opened 2020-01-22
ISSUE-38: Add a Temp file route with IIIF pattern

DiegoPino:
See #38 

# What is new?

We expose now, for logged in users, a direct endpoint to temporary files (e.g uploaded during a Webform ingest/edit workflow). This is needed to allow other modules to show/preview files are not persisted yet and our IIIF server can not reach yet to serve.

This pull also deals with #37 (hopefully) because i'm running out of time to make multiple pulls for everything that requires my attention. Citing myself 
`"Pino, Diego ; mapping: boolean? That was a weird one!"`

### DIFF
--- a/config/schema/format_strawberryfield.schema.yml
+++ b/config/schema/format_strawberryfield.schema.yml
@@ -8,9 +8,10 @@ format_strawberryfield.iiif_settings:
     int_server_url:
       type: string
       label: 'Internal IIIF server URL'
-format_strawberryfield.formatter.*:
-  type: config_entity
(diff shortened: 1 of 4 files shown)


DiegoPino:
Merging this, i know nobody will react on time to review so never mind. We can deal with any issue introduced by this later, if any.

marlo-longley:
@DiegoPino I read the code but didn't test yet. Happy to test as part of overall beta2 testing. I'm excited this brings great possible features in the future.

Issue #40, opened 2020-01-27
Enforce as:media ordering all over the place


DiegoPino:
# What is new?

Strawberryfield Paged Formatter already has a orderPages method based on sequences (Thanks to the wonderful @giancarlobi). Now we need to make that all over the place. Means on the Media Formatter (Openseadragon), Metadata Display Formatter (Injected into a Twig template) and also on our Metadata endpoints (same Twig Template idea).

There are a few ways of doing this, but what i think is the fastest is to add the orderPages method to the base formatter class, and then implement ordering whenever it is needed. But, also, we need to order all as:media types when dealing with Twig templates. For that case i'm adding a constant to base formatter with all the ones we can provide (since those are deduced from the mimetype prefix with an tiny exception...document) and iterate over that ordering. In most cases it should be fast, fast enough. If someone can suggest a way this can be done directly on the twig template, even easier for those cases since that gets cached during render.

Related to https://github.com/esmero/webform_strawberryfield/issues/12
and https://github.com/esmero/strawberryfield/issues/60

marlo-longley:
@DiegoPino - for the direct use on Twig template. If I understand correctly your question? Twig provides a `| sort` filter you can add to a for loop in a template, but it is based on php's `asort` not `uasort`, so we could easily write our own filter.

DiegoPino:
@marlo-longley Twig has recently added arrow functions to sort
https://twig.symfony.com/doc/2.x/filters/sort.html (see the examples)

Means we could get around the usort issue. We don't need natural ordering in the twig template, just that the existing, if any 'sequence' key is used.

We can also add a filter as you suggest, but i feel it could be an overkill? And make our templates way less compatible?

But, here is the catch. Drupal 8.8.1 ships with Twig 1.4.2 but is compatible with Twig 2.x good, sadly `drupal/console-core 1.9.4` is fixed on 1.x! See https://github.com/hechoendrupal/drupal-console-core/blob/master/composer.json#L53

So i would say, lets document the Twig 2.0 option in an issue and leave this as a @TODO once drupal/console-core gets bumped. We can not patch a composer.json file via composer. So we are kinda stuck there. We can always remove drupal console (we are mostly using drush).

marlo-longley:
@DiegoPino ok good point, would be nice to add our own sort arrow function.
I can open an issue with documentation.

DiegoPino:
@marlo-longley thanks

DiegoPino:
[...]

PR #41, opened 2020-01-27
ISSUE-40: Add Media ordering to All formatters that need it

DiegoPino:
See #40 

# What are we adding here?

Sequence ordering of our as:media elements everywhere they are used in output/formatters.

This is based on @giancarlobi work on the `::orderPages` method. It just extends the same concept to every other formatter and Twig injecting process and moved methods to common classes.

@giancarlobi @marlo-longley


marlo-longley:
@DiegoPino looks good, handy use for Base formatter.

DiegoPino:
@marlo-longley detected after merge some bugs related to the fact that not every formatter extends the base formatter (why not also... i mean could we allow people to set also a IIIF server on the Twig template one... we should) so @giancarlobi this pull was wrong and introduced bugs. Fixing now in a new one

PR #43, opened 2020-01-29
ISSUE-40:Fixing media ordering

DiegoPino:
See #41 and #40 

# Bug fix

Pull #41 introduced a bug that escaped my sleepy radar and some other folks radar too. This fixes it.
Requires https://github.com/esmero/strawberryfield/pull/63 to be merged/testing.

basically moves sequence ordering to our JSON Helper class and makes it static. Also renames the constant to uppercase (good practice) and cleans `MetadataExposeDisplayController` from being dependant on a formatter base class, never ever a good idea in the first place.

@marlo-longley please test as soon as possible. This is quite annoying!


DiegoPino:
Now that https://github.com/esmero/strawberryfield/pull/63 is merged (bymyself with code review from @marlo-longley 🎂 ) i can merge the tears away. Thanks!

PR #45, opened 2020-01-30
ISSUE-44 :Fix wrong Metadata Display ID default and check for entity before getting label

DiegoPino:
See #44 (First issue).

The default media entity ID was wrong (long time so). This unsets a default and also does a quick check on an existing entity before trying to set a label in the summary

@marlo-longley ping


DiegoPino:
@marlo-longley testing and seems to work OK. No default does the trick.

marlo-longley:
@DiegoPino this works fine. Same fix I came up with, except my validation was messier, good to know about "validate_reference". 

About this message:
Please configure this formatter by providing one in the configuration form.

I like it and it helps the user know what to do with the form. I thought it was worth pointing out that even if the user makes no selection, the field will work using defaults. But maybe this is more information than they need. Personally it helped me understand the system knowing there was a default formatter that would kick in.

DiegoPino:
@marlo-longley i need to backwards patch beta1 too.

PR #48, opened 2020-02-07
ISSUE-47: Mini-mini-me-malist PDF Viewer  …

DiegoPino:
# Whats this?

Well it took me more hours of course to get this quite basic PDF viewer going.

Since Mozilla encourages (ask please) to people to **NOT use their Viewer directly** without modifying, and re-skinning (and some folks _i know_ have been not reading that well..), lets help Mozilla and go minimal. So i went the minimal, but we can can then extend. Its using PDF.js core only, with async loading of pages and next/prev buttons via a theme hook.

There is quite some work to do to make this look Pro, like when we have more than a single PDF we should show a select button and reload JS based on that.

Also, we need to implement, at least! the Search/highlighter. Option one would be through https://github.com/mozilla/pdf.js/blob/93aa613db7a874e6ed964f394241605d5586c142/web/pdf_find_controller.js but seems quite easier to go reusing the viewer.js[1] component an theming/reducing it, at least simpler than what i did here!

[1] https://github.com/mozilla/pdf.js/blob/master/examples/components/simpleviewer.js#L58

Also:

## Make our Display Mode mappings universally kick-ass  …

Now we enforce our rule (own rules) based rdf/json key type mapper to kick a particulat `View mode` only when looking at an Object in its native/own URL, a.k.a as landing page. 

But hey, what if i remove a condition and this happens everytime the default (named 'full') display mode is present and we actually have a mapping set up? Would that not be great? Well here you have it.

Side note: i really like the function `node_is_page()`, i will forget about his `existence` once i merge this. Live well `node_is_page`. We will meet again some day.

### DIFF
--- a/js/pdfs_strawberry.js
+++ b/js/pdfs_strawberry.js
@@ -1,0 +1,140 @@
+(function ($, Drupal, drupalSettings, pdfjsLib, pdfjsWorker) {
+    'use strict';
+    Drupal.behaviors.format_strawberryfield_pdfjs = {
+        attach: function(context, settings) {
+            // The workerSrc property is a must!
+            // @TODO: Would love to simply push ther 'worker' property since we already have it loaded
+            // but i lack enough understanding of the API.
+            // See here https://github.com/mozilla/pdf.js/blob/master/src/display/api.js#L218
+            // Probably (have to check) passing an Object with that key instead of a string to pdfjsLib.getDocument($doc)
+            //  and worker being the global window['pdfjs-dist/build/pdf.worker'] should work
+            // basically doc.src='url' and doc.worker = window['pdfjs-dist/build/pdf.worker' ?
+            $('.strawberry-document-item[data-iiif-document]').once('attache_pdf')
+                .each(function (index, value) {
+                    var $document = $(this).data("iiif-document");
+                    var $theid = $(this).attr("id");
+                    var $initialpage = $(this).data("iiif-initialpage");
+                    pdfjsLib.GlobalWorkerOptions.workerSrc = '//cdn.jsdelivr.net/npm/pdfjs-dist@2.2.228/build/pdf.worker.min.js';
+
+                    /**
+                     * Single Page render
+                     * Could be used for thumbnails?
+                     */
+                    function singlePageRender() {
+                        // Asynchronous download of PDF
+                        var loadingTask = pdfjsLib.getDocument($document);
+                        loadingTask.promise.then(function (pdf) {
+                            console.log('PDF loaded');
+                            // Fetch the first page
+                            var pageNumber = 1;
+                            pdf.getPage(pageNumber).then(function (page) {
+                                console.log('Page loaded');
+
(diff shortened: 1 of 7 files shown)


src/Plugin/Field/FieldFormatter/StrawberryPdfFormatter.php
```diff
@@ -0,0 +1,303 @@
...
+                // @TODO we can bring a plugin here and there that deals with
+                $elements[$delta]['pdf'.$i] = [
+                  '#markup' => '<i class="fas fa-times-circle"></i>',
+                  '#prefix' => '<span>',
+                  '#suffix' => '</span>',
+                ];
+              }
+            } elseif (isset($mediaitem['url'])) {
+              // TODO. We can serve non mananged by US PDFS directly here
+              // We would just have less data. But that is all
```
DiegoPino: Comment is deceiving. If the remote URL has no CORS enabled, then via JS we have no chance. Lets keep it this way

DiegoPino:
@marlo-longley i'm merging this. The base is here, now i can improve it, but i will do that as a second pass. There are things i would love to discuss and probably pieces missing, so please feel free to review even if its merged. Thanks!

js/pdfs_strawberry.js
```diff
@@ -0,0 +1,140 @@
...
+            // See here https://github.com/mozilla/pdf.js/blob/master/src/display/api.js#L218
+            // Probably (have to check) passing an Object with that key instead of a string to pdfjsLib.getDocument($doc)
+            //  and worker being the global window['pdfjs-dist/build/pdf.worker'] should work
+            // basically doc.src='url' and doc.worker = window['pdfjs-dist/build/pdf.worker' ?
+            $('.strawberry-document-item[data-iiif-document]').once('attache_pdf')
+                .each(function (index, value) {
+                    var $document = $(this).data("iiif-document");
+                    var $theid = $(this).attr("id");
+                    var $initialpage = $(this).data("iiif-initialpage");
+                    pdfjsLib.GlobalWorkerOptions.workerSrc = '//cdn.jsdelivr.net/npm/pdfjs-dist@2.2.228/build/pdf.worker.min.js';
```
marlo-longley: Yes, too bad loading this. Saw the docs but would need to test around like you suggest.
```
The worker shall be built into a separate bundle: take the file "./node_modules/pdfjs-dist/build/pdf.worker.entry.js" or built a separate file that uses require('pdfjs-dist/build/pdf.worker'). PDFJS.workerSrc shall be set to point to this file.
```

DiegoPino: Its a bummer. Also, totally not possible to push. If i preload the worker, it takes (even when not around) a Port and start waiting for work. So when i push it into the workerSrc  (or smartly via property) PDF.js starts to initialize a new instance and fails! So bad and i get a warning. Anyway, this way at least its working (see how i don't load it in the `library` yml file at all because of that). Thanks for checking this. Now we need to implement search, download, and print!

Issue #59, opened 2020-02-25
Normalize Formatter width options


DiegoPino:
# What is needed?

Allow every formatter that renders some type of display (except metadata, that can be for sure be setup directly as css somewhere) to use '0' width to force a 100% width display.

All this can always be overridden by CSS and themes, but its good to give theme-less people also the change to set that up. Right now Mirador is already making use of this feature

DiegoPino:
Solved via https://github.com/esmero/format_strawberryfield/commit/75203881e505643f84986840207aab9711e2bbfe

PR #60, opened 2020-02-26
ISSUE-59: Allow 100% with for all formatters and some ++candy

DiegoPino:
See #59 

# What is new?

All(well almost) formatters now support '0' width option to be used as a stub for 100% width.

The others, like some formatters that depend on Canvases (which need to have a fixed pixel size)e.g 3D one have an extra sizing functionality via JS that contracts the given canvas size in pixel to whatever is smaller (screen smaller than fixed with? We shrink to screen and redraw. Awesome!).
Also, other Formatters that totally need to have control over aspect ratio like PDF and Video allow to set '0' to force an `auto` height on a given `fixed` width. Great for avoiding squeezed video and PDFs like from 1997 when i was alive...and coding..

# What else is new?

So JS improvements, like a `CSS` based preloader for our `3D Viewer` (who wants to wait looking at a white screen for a 3D skull of 10 Mbytes to render? Not me !)

Also all  Formatter summaries in the Display Mode configs reflect this many options. Either `pixel X pixel`, `% x pixel` or `% by auto`. Cool!

We can now render a single frame of 3D into a base64 encoded Image! Well, with some little more code (like an html div and an img src where to put that data, not really code right! We can generate on the fly thumbnails for 3D. (no iiif for that... there is no iiif for 3d, even when there are `3 x is` in iiif)

I also added an extra mime type of `model/obj` Did you kids know Drupal8 ships with not so many useful mime types of us repo users?

I had to rebase a few times so some things could look weird. Will squash when done to cover my strange rebase (if any.. am i a good re-baser?...)

### DIFF
--- a/config/schema/format_strawberryfield.schema.yml
+++ b/config/schema/format_strawberryfield.schema.yml
@@ -142,7 +142,7 @@ field.formatter.settings.strawberry_metadata_formatter:
     metadatadisplayentity_id:
       type: string
     metadatadisplayentity_uselabel:
-      type: string
+      type: integer
 field.formatter.settings.strawberry_paged_formatter:
   type: mapping
   label: 'Specific Config for strawberry_paged_formatter'
@@ -301,7 +301,7 @@ field.formatter.settings.strawberry_mirador_formatter:
     metadataexposeentity_source:
       type: string
       label: 'metadataexpose_entity machine name'
-    manifestnodelist_json_key_source:
+     manifestnodelist_json_key_source:
       type: string
       label: 'Strawberryfield/JSON key containing NODE ids or UUIDs from which to generate Manifest URLs'
     manifesturl_json_key_source:
--- a/config/schema/format_strawberryfield.schema.yml
+++ b/config/schema/format_strawberryfield.schema.yml
@@ -350,4 +350,3 @@ ds.field_plugin.*:
     formatter:
       type: field.formatter.settings.[%parent.%parent.formatter]
       label: "Formatter settings for a generic ds.field plugin"
-
--- a/config/schema/format_strawberryfield.schema.yml
+++ b/config/schema/format_strawberryfield.schema.yml
@@ -142,7 +142,7 @@ field.formatter.settings.strawberry_metadata_formatter:
     metadatadisplayentity_id:
       type: string
     metadatadisplayentity_uselabel:
-      type: integer
+      type: string
 field.formatter.settings.strawberry_paged_formatter:
   type: mapping
   label: 'Specific Config for strawberry_paged_formatter'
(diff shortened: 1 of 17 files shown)


giancarlobi:
@DiegoPino There is a strange issue, probably I'm doing something wrong. Image with Media formatter with IIIF Media and book with Page formatter with IAB when set width to 0 and height > 800 are displayed 723px x height while if I set width 1024 are displayed 1024px x height. 
Are this right? I supposed if width=0 then changing height the width has to change ...

giancarlobi:
@DiegoPino you can check that here: http://archipelago.byterfly.eu/node/25 http://archipelago.byterfly.eu/node/18 bith set to Maximum size: 100% x 900 pixels

DiegoPino:
@giancarlobi thanks for reporting this.

I just checked your site. I see right now this for the Media (openseadragon) container
```html
<div id="iiif-field_descriptive_metadata-d81e698a-a1b5-4c85-84b2-677e990e72da-0-media1" class="strawberry-media-item field-iiif container" data-iiif-infojson="http://archipelago.byterfly.eu/iiif-server/iiif/2/53d%2Fimage-giovane-uomo-del-ballerino-che-indossa-un-salto-russo-piega-del-costume-28730977_3.jpg/info.json" data-iiif-group="iiif-field_descriptive_metadata-d81e698a-a1b5-4c85-84b2-677e990e72da-0-media" data-iiif-thumbnails="1 as:image 900 http://archipelago.byterfly.eu/iiif-server/iiif/2 http://150.145.48.51:8182/iiif/2" style="width:100%; height:900px">
```

And style is correctly 100% and the height 900px.
The center region where this is set is 615px width so it can only reach that width. 

Maybe i'm missing something? Please let me know how this should behave to find the problem and fix it.


Just in case i momentarily disabled JS aggregation in your performance configuration and also hit manually to make sure it refreshed from any caches, so that could have been also the case, all this JS makes browsers quite hungry for caching!

Will look at a book now and report back!

giancarlobi:
@DiegoPino It's working good!!! It was my misunderstanding about how that works. As you suggested, resizing browser window ratio changes according 100% setting. Sorry and thanks.
I'll check now for PDF, I heve to ingest then check viewer 100% setting.

DiegoPino:
Cool. I will also ingest a 3D Model if you are OK with that =)

giancarlobi:
> Cool. I will also ingest a 3D Model if you are OK with that =)

Sure!!!

giancarlobi:
> I'll check now for PDF

@DiegoPino Also PDF works great! http://archipelago.byterfly.eu/node/29 In addition also 100% x auto is great, currently set un object linked above

DiegoPino:
@giancarlobi great! 3D model up and working =)
http://archipelago.byterfly.eu/node/30

I added an PNG image (the texture) to the ADO too. Not used right (only as thumb i guess) but we will be able to allow the 3D to load any as:images as textures too soon =)

DiegoPino:
@giancarlobi we made it! Great, thanks for all the testing. This will make Phone users and theming way easier. I appreciate this

Issue #78, opened 2020-06-23
Allow Metadata to affect Display/Formatter settings


DiegoPino:
# Well what is this heresy? Metadata ruling over how things look?

Like obviously this was needed. I means really folks, how can we expect that a set of fixed settings (hu, 'Viewer hieght') will apply to all objects of a single type? Well not in my watch

This issue (with a commit to get us started) will cover the following:
Add a base method to our formatter base class that will allow any JSON keys/structure the user decides on to feed back/complement a set of predefined, safe, settings on each formatter.  These can range from tiny things (like feed the width from exif) to larger things like totally decide/make some calculations based on multiple Metadata elements on how things should behave.

**First step is easy:** make this happen for the most complex viewer. (not mirador.. for that we have IIIF). Pannellum.
**Second step:** Make it for every other formatter
**Third step:** allow also Twig templates to be used as calculators/inbetweens the data/configs (so you really don't need to hardcode a setting into the JSON, but more like 'calculate it' including basic logic
**Fourth step:** there is no fourth step. C'mon 3 are enough.

@giancarlobi this is the one i told you of. Making the pull now @mitchellkeaney   👀

DiegoPino:
Wonder how we can abstract this functionality in an easy to manage UI way

- We want certain values to be passed back to any viewer. Should viewers have a private method where they advertise what they allow? Most of the time this will end being pushed as JS Object properties.
- We want certain values to be passed as IIIF endpoint values. Rotation/Cropping/Filters


Any other use case?

DiegoPino:
"ap:viewerhints". We have this

PR #79, opened 2020-06-23
ISSUE-78: Metadata can now rule over display

DiegoPino:
See #78 

This pull will also solve a few other tiny things, like adding/pushing other fun things into the twig templates (hey, hey!) and wait for it... 2 more options for the static image formatter (nice one that we can use for thumbnails and such)..
- No image? Nothing to see? Choose a FONT! yep. a mapping per type: For now a static set using Fontawesome, but i expect this to have its own Config Form (who said me?)
- No image but part of another object? Or child of another one? or the opposite? Dad? Yes. Get your images from your neighbors' front yard and reuse. Plan here is to make that very action compatible, method for other stuff. Like 'compounding' and ACL inheritance. But that comes later...

### DIFF
--- a/config/schema/format_strawberryfield.schema.yml
+++ b/config/schema/format_strawberryfield.schema.yml
@@ -200,6 +200,8 @@ field.formatter.settings.strawberry_pannellum_formatter:
       type: string
     json_key_hotspots:
       type: string
+    json_key_settings:
+      type: string
     json_key_multiscene:
       type: string
     panorama_type:
--- a/format_strawberryfield.libraries.yml
+++ b/format_strawberryfield.libraries.yml
@@ -63,12 +63,12 @@ iiif_iabookreader_strawberry:
     - core/drupal.debounce
     - format_strawberryfield/iiif_iabookreader
 pannellum:
-  version: 2.5.4
+  version: 2.5.6
   css:
     component:
-      https://cdn.jsdelivr.net/npm/pannellum@2.5.4/build/pannellum.css: { external: true}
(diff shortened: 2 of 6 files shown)


DiegoPino:
To test:

get yourself a panorama from esie.space

Now enable configure your Panorama formatter to a JSON keys for settings

and add something like this to your JSON via the RAW editor (don't copy and paste, read first and check the comment i added which github conveniently for me marked in red!)
```JSON 
"as:formatter": {
            "urn:uuid:5616e42e-0a0b-4369-860b-9752d6bd5409": {
                "dr:fid": 39, ---> replace this by the image you want to target.. what? One PER IMAGE! yeah, cool
                "settings": {
                    "haov": 149.87,
                    "vaov": 54.15,
                    "maxYaw": 76.935,
                    "minYaw": -76.935,
                    "vOffset": 1.17,
                    "maxPitch": 30.245,
                    "minPitch": -27.905
                }
            }
        }
    },
````

DiegoPino:
Now looking at my own code i figure out i should also allow general settings. Not only ones that apply to a single image. Ideas? How to make the same happen without a urn:uuid:5616e42e-0a0b-4369-860b-9752d6bd5409?
Like with an urn:uuid:* ? That would be almost a regular expression-metadata-key!

DiegoPino:
@giancarlobi i'm merging this in the meantime but will keep this issue open. Because i will add a few ISSUE-78b and c branches as i code more and make the metadata to display effect better!

Issue #74, opened 2020-03-31
Allow IIIF Manifests without an Image Service or with static images to work on IABookreader/Mirador


DiegoPino:
Related to #71 

# What is needed? 

Now that PDF serving via IIIF is fixed on IABookreader we need more!

Given the fact that on JS we work on a Browser/HTML/CSS World, we need to know the width and height of the images we serve via IIIF always, to scale things and display them correctly. In the absence of an Image Resource Service definition in a IIIF Manifest, we have no access to a info.json, which means we can not get the size.

Solution is to either push size/width for every image from the SBF JSON into the IIIF Manifest when creating the Twig Template (we need SBF Runners for that) but in the case of a PDF that is serving pages via an Cantaloupe argument, that is not possible.

Solution: use JS!

https://github.com/esmero/format_strawberryfield/pull/73#issuecomment-606160883

This will need to be either ported to Mirador 3 too, see my comment at the end of 

https://github.com/ProjectMirador/mirador/issues/2616

@giancarlobi

DiegoPino:
@giancarlobi i was thinking loud there, that maybe a work around is to have an internal, Archipelago Only info.json endpoint for this cases. I don't think we can expect our use case to be fixed on any other viewer, but i do want to have this feature.

So, if i can expose an info.json endpoint for our own files, like under https://github.com/esmero/format_strawberryfield/blob/master/format_strawberryfield.routing.yml#L53

The trick here would the following, We do return the same info.json as cantaloupe (if there is one) but we change the base URI to our own wrapper that uses ?arguments as part of the ID? and we do allow/pass around the GET arguments, and only the ones we trust. What do you think? Too crazy?

giancarlobi:
@DiegoPino I fully agree we don't have to wait a fix and yes we want this feature.
We have manifest (of PDF) with:
A) Resource -> @ID : URL with ?page that is working with Cantaloupe
B) Resource -> Service -> @ID : URL with ?page DOESN'T return info.json
So we have to solve B) issue by:
1) modify manifest Resource -> Service -> @ID with an URL pointing to our Archipelago custom endpoint
2) expose Archipelago custom endpoint to return right info.json for PDF paged object.
Did I understand your idea? I think is a good solution!

DiegoPino:
@giancarlobi yes! That is the idea.  for PDF and Video (where the argument is a timecode)

DiegoPino:
@giancarlobi also 👀 
[...]

PR #83, opened 2020-07-15
Issue 74b: IABookreader, please cooperate with old friend PDF?

DiegoPino:
See #74 

# What is this?

Some JS magic to deal with missing/partial/missunderstood Manifests.  It does the following
- If service level is > 0 (means we have an info.json), append the get arguments to it (?page=1)
- If width/height is present but level == 0 (like static image or one we want o set because we are like that) then use the Canvas width/height ratio to calculate the missing one. (either height or width)
- If service level == 0 don't even bother trying to figure out an .info.json and in the worst possible condition just use the generic 3/4 ration and console it out so the webmaster/admin/can have nightmares and fix its Twig Template.

Gosh! @giancarlobi i know you out but someone needs to see this even if in 2 days. @alliomeria i need to introduce you IIIF Manifests sooner than later.

In conclusion. This allows Bookreader to display individual pages for PDFs without any issue, even if we don't want to fake the size

### DIFF
--- a/js/mirador_strawberry.js
+++ b/js/mirador_strawberry.js
@@ -43,7 +43,7 @@
                         }
                         //@TODO add an extra Manifests key with every other one so people can select the others.
                         var miradorInstance = Mirador.viewer($options);
-                        console.log('initializing Mirador 3.0.0-beta.4')
+                        console.log('initializing Mirador 3.0.0-RC3')
                     }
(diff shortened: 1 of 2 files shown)


DiegoPino:
Merging folks. No time to loose. Also going to bed. Have been up more than it's legal 😴

PR #88, opened 2020-09-02
Reflect 1.0.0 rc1 and  Comment out duplicated _entity_access 

giancarlobi:
# What is the issue?

Drupal allows certain shortcut keys in its ROUTE definition YAML files. One of those is `_entity_access` which allows to enforce that if an argument in the route is a given entity, one of the permissions against that entity needs to return `Allow` before calling the controller. Previously (because of an undocumented code bug) we could for this particular Route that has two entities in its path (a node and a metadata display) use the argument twice. 

Now it is back to strict. The long term solution here, to ensure that both Node and Metadata Display entity access is checked before calling the controller, is to create a custom access check that gets both arguments and does an `AndIf` comparison of the  permissions applied to each entity. For now the fix to make this work on 9.x is good since Twig templates are `metadata display` not really being enforced to be hidden in any of our implementations and the Controller itself is doing an extra final check on access anyways. 

Finally this pull @giancarlobi  is making also adds as dependency Webform 6.0 which is Drupal 9 only but keeps 5.19 if on 8.9.x

PS: Thanks Giancarlo!


DiegoPino:
@giancarlobi can i add a description to your pull? Or if you can? Just want to be sure i remember in the future what this was about/why. Thanks!!

giancarlobi:
> @giancarlobi can i add a description to your pull? Or if you can? Just want to be sure i remember in the future what this was about/why. Thanks!!

@DiegoPino Of course, better if you add it because you know better the issue

DiegoPino:
@giancarlobi i will also open an issue to fix long term and make that access check a custom one.

PR #89, opened 2020-09-02
Updates JS to comply with Drupal 9 missing attachments

DiegoPino:
# What is this?

@giancarlobi Seven Theme (in Drupal 9) does not include modernizr anymore. We use it for the bookreader. Here is the patch!


DiegoPino:
@giancarlobi cool. Now your other pull became a commit here. I will not squash but merge so both commits get into the History (since you are making history! 🥇 )

DiegoPino:
Removed all reviewers since it is just you and i and both added code. Merging!

Issue #91, opened 2020-09-08
Test, fix and update JS libraries for 1.0.0RC1 and D9 compatiblity


DiegoPino:
# What is this?

A global large-ish issue so i have a number to use for the upcoming branch name that will add a few commits, in the form of bumping some versions of JS, adding a few more (Universal viewer) and a hook (experimental) that with a switch of a key (not really, more like a setting form) will allow Archipelagos to use (and download hu!!) local JS versions too in case you need to run your server from inside a public bathroom in the middle of the moon without Internet access.

PR #92, opened 2020-09-08
Issue-91: update JS libraries/Twig/Formatters and prepare of D9 deployment

DiegoPino:
# What is new?

See #91 
- Minor JS updates including most of our delivered viewers and libraries
- Updates to references to remote Images and assets
- Added Leaflet Marker Cluster, an old friend of mine (from back at Chile times) for larger maps
- Universal Viewer Core Library soon to be complemented with the rest of the framing, our own JS and the Formatter code
- A Twig extension to check the type (class) of an object, useful if you want to call NODE or User methods on unknown classes of Objects
- A Twig deprecation See #94 
- W3C Web Annotations with "adding", removing and updating directly on the Open Seadragon Viewer !(gosh am I good! hahaha) Tested and works. This also adds some new permissions and an almost full blown CRUD/REST API for the web annotations. Even work (designed so) with multi image slides. Annotations are keyed by Image URL. Not happy? you can always discard them before saving.

 @giancarlobi ready!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,8 @@
     "require": {
         "ml/json-ld": "^1.2",
         "mtdowling/jmespath.php":"^2.5",
-        "strawberryfield/strawberryfield":"dev-8.x-1.0-beta3",
-        "drupal/webform": "^5.19",
+        "strawberryfield/strawberryfield":"dev-1.0.0-RC1",
+        "drupal/webform": "^5.19 || ^6.0",
         "drupal/webform_views": "^5.0"
     }
 }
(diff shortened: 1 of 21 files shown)


giancarlobi:
@DiegoPino the first couple of commits are ok.

DiegoPino:
@giancarlobi thanks. I want to group a few commits together here since opening too many issues and making small changes is becoming complex to sustain, review and test wise. Universal viewer seems to be a good solution also for .epub, Video, Audio, in the IIIF V3 environment so would love to have at least the base formatter working on D8 and D9

DiegoPino:
@giancarlobi @alliomeria W3C Annotations are ready. Double tested, caching issue solved, Delete works perfect, CRUD works well now, Discard Reloads the Edit page so we get our old ones back. All good. Merging into RC1!

PR #93, opened 2020-09-11
Make IAB ZOOM working with IIIF PDF

giancarlobi:
The main issue here is preserve query string in Cantaloupe request so we override a couple of openseadragon functions. Also add OSD override library to IAB viewer as ZOOM is needed when we use IAB.


DiegoPino:
@giancarlobi this is amazing!!

this also opens new chances for OSD to be used directly as viewer for PDFs/Or even Movie Frames without the need of IABook reader!!

DiegoPino:
@giancarlobi will test in my local instance with our demo objects and will merge ASAP. Thanks a lot. Great contribution

DiegoPino (approved):
Good, works great! FYI: Only change i will probably do later before release is to make Book reader completely depend (`dependencies:`) on our openseadragon library instead on the js/iiif-openseadragon_strawberry.js file one. That way the work on Web Annotations can be also used on the bookreader, etc. Thanks @giancarlobi you fixes this good!! (JS ace)

Issue #100, opened 2020-11-05
Metadatadisplay/Twig render enhancements (for powerpuff users)


DiegoPino:
# What is up Bubbles? 
_(Dad joke because render caching `bubbles` up  cache context during a twig template rendering.. too much explanations for a bad bad Drupal joke..gosh)_

So: What happens if you need to add extra static data to your Twig template/Metadata displays? Data that is not in your ADO metadata. Let's say you want to have the full `ISO693-2` Language and code list (with original language too) around so you can fill up via Twig template a quite complex EAD2002 XML finding aid? And, again, you have `none` of that data in the actual SBF JSON data (the ADO's metadata?).

You would most likely blame `Mojo Jojo` but you can also let poor Mojo in peace and think and deduce that `Metadata Display entities` are actual `Content Entities` and are Field-able. So what this issue adds, is a -if you add a Strawberry Field == JSON type of Field- to your Metadata Display Entity (name it whatever you want), now you will have during `edit` an extra Field (text area) to fill? Think `Bubbles`, what can we put there?
- STATIC JSON that will be injected and be available (superpower) as data during render time. 

Humble example here:
![image](https://user-images.githubusercontent.com/6946023/98303780-b5bee080-1f8c-11eb-9778-8a9789373bc6.png)

This depicts a fresh and silly added Strawberry field named extra_data. With a JSON that contains all ISO693-2 codes.

Once saved the twig template has access inside a `datafixed` key (additional to the dynamic `data` and other context elements (@alliomeria we need to document but also add as help in the edit form)

CONS? 
- Its an additional SQL query (only one.. and not that terrible). So not needed for every instance and the extra field will not be shipped by default. If not present/no extra query
- Requires the user to understand how things work. So `powerpuff` type of user
- WE only allow this for SBF fields. If you add String? Good for you. But won't be passed to the Twig `Context` as a data structure
PROS? 
- Extra fixed data, hashes, mappings, vocabularies, etc. FIXED. This is not meant for dynamic data.

Additional. I noticed that our twig template rendering had mixed approaches and our entity itself has already self rendering methods used in the endpoints, so that will be also updated. Both formatters that use inline rendering will be calling the right methods and using the cache invalidation correctly.

@giancarlobi @alliomeria we may need some testing here. This is advance puff-stuff

![Advanced Testing in rabbit costume is demanded](https://media.giphy.com/media/SNIBaECFzybug/source.gif)

DiegoPino:
[...]

PR #101, opened 2020-11-05
ISSUE-100: Allow extra data to be pushed into a Metadata Display via its own custom fields

DiegoPino:
Tired, so just read #100, all the cuteness and jokes are there!

@giancarlobi @alliomeria we need to test this thoroughly. There are CORE changes and improvements added to rendering Twig templates but I want to be sure they work correctly in every possible case. Should not break (tasted here of course) and the BIG improvement of course is "optional", not everyone will need to push extra Data into a Twig template.

This also proofs (yeah) that Strawberryfields are awesome and can be used in other entities too.  I also want a piece of cake because I managed to remove Code from Archipelago making everything more slick and cute.

### DIFF
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -1,5 +1,7 @@
 <?php
 
+
+
 namespace Drupal\format_strawberryfield\Entity;
 
 use Twig\Node\ModuleNode;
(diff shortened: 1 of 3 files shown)


DiegoPino:
Giancarlo did some Apache benchmark test and there is a slight increase (milliseconds) in first delivery time, that then normalizes completely after caches kick in. I feel this is good and we can iterate back into performance testing if it becomes annoying. Sometimes Super fast can be super slightly less super fast if we want to give users more flexibility. And we do that here.

PR #104, opened 2020-11-19
ISSUE-102: large files via Ranged request delivery, replay.web 1.3.0

DiegoPino:
# What is new?

Won't repeat. Its all here https://github.com/esmero/archipelago-deployment/issues/75#issue-745187010
Gist: large files leak memory when delivered via Range requests because Guzzle/AWS SDK and Symfony have issues 🤕 
Solution: we make the same requests we get back to the storage and only stream the bytes we need without offset on the PHP machinery. Sleek, elegant, hack-ish, perfect. Works

This also includes the new replay web 1.3.0 JS and Mirador 3.0.0 final.

Comes with bonus of a new version of PHP in our docker containers. 7.4.9. Nice and wazc tool to transform WARC files into WACZ files

### DIFF
--- a/format_strawberryfield.libraries.yml
+++ b/format_strawberryfield.libraries.yml
@@ -93,6 +93,7 @@ av_strawberry:
   css:
     component:
       css/htmlaudiovideo.css: {}
+
 jsm_modeler:
   version: 0.45
   js:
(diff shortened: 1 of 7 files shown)


giancarlobi:
@DiegoPino really a great job!!! Congrats friend!! If you need some more checks onto my Archipelago (filesystem) and PHP 7.3.24-3+ubuntu18.04.1+deb.sury.org+1 (update PHP, check transf big file, ...) I'm here.

PR #111, opened 2020-12-20
Fix broken IAB Zoom on PDF source

giancarlobi:
@DiegoPino I found this was broken after last PR, this patch fix it, please check if ok also for you. I added a your function at the end because not available in this js file, I think could be optimized.

### DIFF
--- a/js/iiif-iabookreader_strawberry.js
+++ b/js/iiif-iabookreader_strawberry.js
@@ -137,7 +137,8 @@ BookReader.prototype.buildViewpageDiv = function(jViewpageDiv) {
   if (1 == this.mode) {
     var index = this.currentIndex();
     //OLD//var tilesourceUri = this.getPageURI(index, 1, 0).replace(/full.*/, "info.json");
-    var tilesourceUri = this.getPageProp(index, 'infojson');
+    //var tilesourceUri = this.getPageProp(index, 'infojson');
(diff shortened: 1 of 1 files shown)


DiegoPino:
@giancarlobi great! Can you explain what was wrong? I also saw it broken but was not clear on the issue. Also, what happens if the manifest is not using FULL (just a doubt because I do not understand it yet but I totally trust you on this). What about the other pull and the padding? Thanks for fixing this friend!!

giancarlobi:
Dear friend, I'm just writing code (a draft) for pdf searchable and how to check and select between tesseract and djvu2hocr, I hope (if my neurons kept alive) to end in this night.
With ZOOM I found that after last PR, when source are PDFs, the value returned here getPageProp(indices[0], 'infojson') is NULL while if source is TIFF that doesn't happen. So I revert to old style by getPageURI + query params due to pdf page query.
About if manifest doesn't use FULL we have to check, I feel it has to work anyway, to check
About padding, I saw that if increase of 1 (yes, one) px the container then the last page can go up how much needed to nav bar to recognize that changed, a silly math rounding issue, I suppose.

DiegoPino:
Perfect. Makes sense. I will merge during my afternoon and will check with other manifests later tonight! Thanks!!!

giancarlobi:
> Perfect. Makes sense. I will merge during my afternoon and will check with other manifests later tonight! Thanks!!!

It's Sunday, friend, so no rush and merge and check when you have free time, please.

DiegoPino:
Sunday who? hahaha

PR #119, opened 2021-01-28
ISSUE-118: Adds a way to hide the search bar in the paged formatter if the document is not processed via ocr. 

### LINKED ISSUE #118: Hide search input in the Paged Field Formatter / Internet Archive Book reader if no OCR
# What is this?

Now that we have OCR we want to avoid confusing users and leading them to think they can "search" if no Solr Document with OCR exists.

How to do this?
1. Find out if a certain ADO (Node bearing a strawberryfield) had already `strawberryfield_flavor_datasource` search api sources documents generated and those can be found.
  - Look at this code:
  - https://github.com/esmero/strawberry_runners/blob/main/src/Plugin/QueueWorker/AbstractPostProcessorQueueWorker.php#L530
  This method can be copied/simplified and abstracted/generalized to be able to check (true/false) if any ADO has a particular post processor generated Search API index entry. The way I see this is, abstracting/copying this method to be check for "any" instead of a specific one Search API ID that matches the critters and be used in a service either in this `format_strawberryfield` module or in any of the utility Service that exists in our core `strawberryfield` module (e.g https://github.com/esmero/strawberryfield/blob/1.0.0-RC1/src/StrawberryfieldUtilityService.php)
In other words. The function should have as arguments: the NODE ID, and the PostProcessor Plugin ID we want (in this case `ocr`)
which Search API fields to target can be seen in the Search API/ SOLR Fields tab for the Strawberryfield flavor datasource via the UI, but also, this is how one of those Solr docs look like (I removed all `sort_*` fields here to make it more compact)
```JSON
      {
        "id":"5qc1nk-default_solr_index-strawberryfield_flavor_datasource/4:1:en:6ba944b8-8518-415e-abd2-4af808a457c7:ocr",
        "sm_context_tags":["search_api_X2f_index_X3a_default_solr_index",
          "search_api_solr_X2f_site_hash_X3a_5qc1nk",
          "drupal_X2f_langcode_X3a_und"],
          "tm_X3b_und_fulltext_with_original_ocr":["<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<ocr><p xml:id=\"sequence_1\" wh=\"2550 3300\"><b><l><w x=\".118 .095 .085 .014\">Sheppard</w><w x=\".209 .095 .052 .011\">Helen</w><w x=\".267 .095 .104 .01\">...."],
        "ss_search_api_language":"und",
        "ss_title_2":"Helen Sheppard Oral History Interview (Excerpt)",
        "sort_X3b_en_processor_id":"F.L\u0000",
        "timestamp":"2021-01-13T17:14:44Z",
        "site":"http://localhost:8001/",
        "ss_parent_id":"4",
        "ss_sequence_id":"1",   
        "boost_document":1.0,
        "ss_search_api_id":"strawberryfield_flavor_datasource/4:1:en:6ba944b8-8518-415e-abd2-4af808a457c7:ocr",
        "ss_processor_id":"ocr",
        "its_uid_2":0,
        "_version_":1688792554279010304,
        "ss_checksum":"2feffd1c3f381a542db82931982dd834",
        "ss_search_api_datasource":"strawberryfield_flavor_datasource"
}
````

2. Call this new `method` from here https://github.com/esmero/format_strawberryfield/blob/1.0.0-RC1/src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php#L322
If result  `== FALSE` add a Drupal Setting entry/flag that allows us..(continued..)

3. ... In the JS to disable the search box https://github.com/esmero/format_strawberryfield/blob/1.0.0-RC1/js/iiif-iabookreader_strawberry.js#L37, somewhere there. There are [IABOOK reader specific arguments](https://github.com/internetarchive/bookreader) that can be used to disable functionality (that should all the code that is needed on the JS)

4. Make sure we have the right caching context so the FALSE does not get cached forever and the next time a user comes back and the documents are already there the search box is showed

@pcambra this one is for you! Feel free to ask/comment and reach out to me any time.

### PR DESCRIPTION
**Pull request by pcambra**
resolves #118

### DIFF
--- a/src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
@@ -10,31 +10,25 @@ namespace Drupal\format_strawberryfield\Plugin\Field\FieldFormatter;
 
 use Drupal\Core\Field\FieldItemListInterface;
 use Drupal\Core\Field\FieldItemInterface;
-use Drupal\strawberryfield\Tools\Ocfl\OcflHelper;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Component\Utility\UrlHelper;
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\strawberryfield\StrawberryfieldUtilityServiceInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Drupal\Core\Template\TwigEnvironment;
 use Drupal\Core\Field\FieldDefinitionInterface;
 use Drupal\strawberryfield\Tools\StrawberryfieldJsonHelper;
-use stdClass;
-use Twig_Error_Syntax;
-use Twig_Environment;
-use Twig_Error_Runtime;
-use Twig_Loader_Array;
 
 /**
- * Simplistic Strawberry Field formatter.
+ * Strawberry Field Paged Formatter using IABook Readerplugin.
  *
  * @FieldFormatter(
  *   id = "strawberry_paged_formatter",
- *   label = @Translation("Strawberry Field Paged Formatter using IABook Reader
- *   plugin"), class =
- *   "\Drupal\format_strawberryfield\Plugin\Field\FieldFormatter\StrawberryPagedFormatter",
+ *   label = @Translation("Strawberry Field Paged Formatter using IABook Readerplugin"),
+ *   class = "\Drupal\format_strawberryfield\Plugin\Field\FieldFormatter\StrawberryPagedFormatter",
  *   field_types = {
  *     "strawberryfield_field"
  *   },
@@ -52,7 +46,6 @@ class StrawberryPagedFormatter extends StrawberryBaseFormatter implements Contai
    */
   protected $currentUser;
 
-
   /**
    * The entity manager.
    *
@@ -66,25 +59,12 @@ class StrawberryPagedFormatter extends StrawberryBaseFormatter implements Contai
   protected $twig;
 
   /**
-   * Constructs a FormatterBase object.
+   * The Strawberry Field Utility Service.
    *
-   * @param string $plugin_id
-   *   The plugin_id for the formatter.
-   * @param mixed $plugin_definition
-   *   The plugin implementation definition.
-   * @param \Drupal\Core\Field\FieldDefinitionInterface $field_definition
-   *   The definition of the field to which the formatter is associated.
-   * @param array $settings
-   *   The formatter settings.
-   * @param string $label
-   *   The formatter label display setting.
-   * @param string $view_mode
-   *   The view mode.
-   * @param array $third_party_settings
-   *   Any third party settings.
-   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
-   *   The config factory.
+   * @var \Drupal\strawberryfield\StrawberryfieldUtilityService
    */
+  protected $strawberryFieldUtility;
+
   /**
    * StrawberryMetadataTwigFormatter constructor.
    *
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php [hunk 1]
```diff
@@ -186,98 +140,94 @@ public function settingsForm(array $form, FormStateInterface $form_state) {
...
     if ($this->getSetting('metadatadisplayentity_source')) {
-      $entity = $this->entityTypeManager->getStorage(
-        'metadatadisplay_entity'
-      )->load($this->getSetting('metadatadisplayentity_source'));
+      $entity = $this->entityTypeManager
+        ->getStorage('metadatadisplay_entity')
+        ->load($this->getSetting('metadatadisplayentity_source'));
     }
 
     return [
```
DiegoPino, line 148: Nice. Old code, no DCS. Thanks

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -286,12 +236,9 @@ public function settingsSummary() {
         case 'metadatadisplayentity':
           $entity = NULL;
           if ($this->getSetting('metadatadisplayentity_source')) {
-            $entity = $this->entityTypeManager->getStorage(
-              'metadatadisplay_entity'
-            )->load($this->getSetting('metadatadisplayentity_source'));
+            $entity = $this->entityTypeManager->getStorage('metadatadisplay_entity')->load($this->getSetting('metadatadisplayentity_source'));
```
DiegoPino, line 239: For readability, I prefer methods for long lines (over 80 chars) to be split amongst multiple lines. But that said, I'm also fine. MY IDE (even when using DCS and Symfony ones) breaks them automatically. So if you see me breaking them and you are not into that let me know!

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -338,11 +281,7 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
     // To generate an on the Fly Manifest. We coded our JS to read from manifests
     // Finally we allow also an Manifest URL to be passed.
 
-
-    $nodeuuid = $items->getEntity()->uuid();
```
DiegoPino, line 342: Good. Old code. legacy. Thanks for the cleanup

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -358,8 +297,7 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
       if ($json_error != JSON_ERROR_NONE) {
         return $elements[$delta] = ['#markup' => $this->t('ERROR')];
       }
-      // A rendered Manifest
-      $manifest = '';
```
DiegoPino, line 362: same, old code. thx

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -562,14 +496,23 @@ public function processElementforManifestURL(
           $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]['nodeuuid'] = $nodeuuid;
           $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]['manifesturl'] = $manifest_url;
           $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]['width'] = $max_width_css;
-          $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]['height'] = max(
-            $max_height,
-            520
-          );
+          $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]['height'] = max($max_height, 520);
+          // @see self::processElementforMetadatadisplays()
+          $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]['enableSearchArchipelago'] = $this->searchEnabled($item);
```
DiegoPino, line 501: To spare us from some duplicated code, what about (looking at this now for the first time with 2021 eyes) we make a check on the caller method (Like line 334, inside the $delta loop) and check 
```PHP
if (isset($element[$delta]['media'])) {
 // all the common settings?
}
```
what do you think? I mean since you have done so much good work cleaning this old code, maybe that would save is a few lines.

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -562,14 +496,23 @@ public function processElementforManifestURL(
...
     return $element;
   }
 
+  /**
+   * Returns whether the entity is indexed in Solr and processed by OCR.
+   *
+   * @return bool
+   *  TRUE if the entity is found processed, FALSE otherwise
+   */
+  protected function searchEnabled(FieldItemInterface $item): bool {
```
DiegoPino, line 514: Lovely.

DiegoPino:
Note: this pull depends on https://github.com/esmero/strawberryfield/pull/142 and that one needs to be merged first

DiegoPino (commented):
Just a few comments. This is nice!

DiegoPino:
@pcambra one thing I forgot (and I sometimes also neglect so will be watching myself closer) For Pull requests that come from a certain ISSUE we add to the title the issue
In this case it would be 

`ISSUE-118:the full title`. Makes tracing back in the list (and making release notes for me) easier

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -338,11 +281,7 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
     // To generate an on the Fly Manifest. We coded our JS to read from manifests
     // Finally we allow also an Manifest URL to be passed.
 
-
-    $nodeuuid = $items->getEntity()->uuid();
-    $nodeid = $items->getEntity()->id();
-    $fieldname = $items->getName();
-    /* @var FieldItemInterface $item */
+    /* @var \Drupal\Core\Field\FieldItemInterface $item */
```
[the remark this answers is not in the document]
pcambra: According to Drupal coding standards in https://www.drupal.org/docs/develop/standards/api-documentation-and-comment-standards#var and https://www.drupal.org/docs/develop/standards/api-documentation-and-comment-standards#types, it seems the suggestion is `Always prefix types with the fully-qualified namespace for classes and interfaces (beginning with a backslash). If the class/interface is in the global namespace, prefix by a backslash.`

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
pcambra: I used to do something like `$js_settings = &$element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]` but I've found is not the most stable or secure pattern as it can get easily polluted.

What I'll do is to transform this in a single array declaration which looks nicer, but we might want to consider having a base method with sensible defaults that is then called by the manifest and manifest url methods as there's a fair amount of code shared between both paths. I'd do in a followup, though

pcambra:
I've force pushed after rebasing to add the issue # on the commit message

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -490,15 +430,17 @@ public function processElementforMetadatadisplays(
...
-        $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]['nodeuuid'] = $nodeuuid;
-        $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]['manifest'] = json_decode(
-          $manifest
-        );
-        $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]['width'] = $max_width_css;
-        $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid]['height'] = max(
-          $max_height,
-          520
-        );
+        $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid] = [
```
DiegoPino, line 433: yep. Easier to read

DiegoPino (approved):
Good. now that https://github.com/esmero/strawberryfield/commit/d1200a08a81024966326e9a51b552a7b9a91fd99 is in, we can merge this. Looks cleaner. My JS indentations are quite different so may have to look at that.

DiegoPino:
@pcambra @giancarlobi merging this. Its easier to test all integrated in a RC2 if we need to fine tune or make amends.  Will squash since this branch brings some older commits with it.

Issue #120, opened 2021-01-29
Redirect Admin users to the Live View Mode editing when pressing "Manage Displays" tab + Bonus admin hint


DiegoPino:
# What is needed?

This is a fun and very useful one. I will explain in detail the need.

Archipelago uses conditional `View Modes` using a configuration entity/hook to select dynamically which View Mode is going to be applied to render a given NODE/ADO. The decision is based on the JSON "type" key value and only applies if the View Mode is set to "default" and not being overridden by the user creating/editing the NODE.

This happens specifically here and the code is quite verbose and simple
https://github.com/esmero/format_strawberryfield/blob/1.0.0-RC2/format_strawberryfield.module#L95-L134

This brings to Admin level UX issues:
1. When you press the "Manage Display Tab" as an admin your blind to which of the many displays listed is the one actually affecting the rendered ADO. It defaults to the actually "saved" View Mode but not the one we select Live. A lot of guessing will happen.
2. When looking at the Rendered Object also, as an admin, it would like to know Why the Object is being rendered as it is. So, a special (Admin Hint) type of permission (a new one, we could name them Admin Archipelago helper permissions or such) could be a simple solution. Roles with that permission could get `Drupal Messages` send to screen giving them hints. These hints can be quite a cool way of teaching also users how things are being rendered and what generates the viewable output so they can Directly edit/affect the sources of those.  Initially I would simply go for a notification with the actual View Mode label and some text. But then, I would expand that also to related resources. E.g, if a certain SBF formatter is using in its config a given Twig template (Metadata Display Entity) we could show a Link to that Metadata Display entity (if the user has access to edit those of course) or just the Label.

Screenshot shows what a user coming from a Book using during render the `Digital Object with Book Reader` sees. It sees the Default one. Basically making every next operation a multi Browser tab and forcing it also to check `http://localhost:8001/admin/config/archipelago/viewmode_mapping` and based on the RAW JSON make some pretty hard brain connections to understand what/why the ADOs is rendered using a book reader.

![image](https://user-images.githubusercontent.com/6946023/106305512-ca78e080-622a-11eb-96b0-d522d8570b25.png)

@pcambra this one is for you! While i review your SBF code.

Please feel free to ask/request more info/comment or fight me back! @alliomeria please give us more feedback on what other info/direct links on render for certain admin users could be useful to have.

Note: Point 2. Will make display for those users slower. But we will gain a lot from this and teaching archipelago will be easier!

[...]

PR #122, opened 2021-02-05
ISSUE-120: Added an admin hint block and tab for the ADO view mode display

pcambra:
This PR adds the following:

- New block `DisplayInformationBlock` that shows the view mode selected linked to the display settings of that view mode
- A local task plugin to inject the right parameters to the Display settings tab per ADO (`ViewModeLocalTask`)
- A consolidated service: `ViewModeResolver` that is called in view_mode_alter but also in the informative block, it does the heavy lifting of selecting the view mode, caching the biggest and common part (getADOTypes)

### DIFF
--- a/format_strawberryfield.module
+++ b/format_strawberryfield.module
@@ -4,14 +4,11 @@
  * Contains formater_strawberryfield.module.
  */
 
-use Drupal\Component\Serialization\Json;
-use \Drupal\Core\Entity\Display\EntityViewDisplayInterface;
-use Drupal\Core\Entity\EntityInterface;
-use Drupal\Core\Url;
 use Drupal\format_strawberryfield\Controller\WebAnnotationController;
 use Drupal\node\NodeInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Swaggest\JsonDiff\JsonDiff;
+
 /**
  * Implements hook_page_attachments().
  *
@@ -36,15 +33,14 @@ function format_strawberryfield_page_attachments(array &$page) {
   $page['#attached']['library'][] = 'core/jquery';
   $page['#attached']['library'][] = 'core/drupal.dialog.ajax';
   $page['#attached']['library'][] = 'core/drupal.ajax';
-
 }
 
+/**
+ * Implements hook_form_FORM_ID_alter() for node_form.
+ */
 function format_strawberryfield_form_node_form_alter(&$form, FormStateInterface $form_state) {
-
   // Deal with Possible Addded Annotations.
-
   if ($form_state->get('hadAnnotations')) {
-    $node = $form_state->getFormObject()->getEntity();
     $count = $form_state->get('hadAnnotations');
 
     $form['annotations'] = [
@@ -90,47 +86,22 @@ function format_strawberryfield_form_node_form_alter(&$form, FormStateInterface 
   }
 }
 
-
-
-function format_strawberryfield_entity_view_mode_alter(&$view_mode, EntityInterface $entity, $context) {
-  if ($entity->getEntityTypeId() == 'node' && $view_mode == 'full') {
-    $adotype = [];
-    $original_view_mode = $view_mode;
-    /* @var \Drupal\Core\Config\ImmutableConfig $config */
-    if ($sbf_fields = \Drupal::service('strawberryfield.utility')->bearsStrawberryfield($entity)) {
-      foreach ($sbf_fields as $field_name) {
-        /* @var $field StrawberryFieldItem */
-        $field = $entity->get($field_name);
-        if (!$field->isEmpty()) {
-          foreach ($field->getIterator() as $delta => $itemfield) {
-            /** @var $itemfield \Drupal\strawberryfield\Plugin\Field\FieldType\StrawberryFieldItem */
-            $flatvalues = (array) $itemfield->provideFlatten();
-            if (isset($flatvalues['type'])) {
-              $adotype = array_merge($adotype, (array) $flatvalues['type']);
-            }
-          }
-        }
-      }
-    }
-    if (!empty($adotype)) {
-      $config = \drupal::config(
-        'format_strawberryfield.viewmodemapping_settings'
-      );
-      $viewmodemappings = $config->get('type_to_viewmode');
-      $viewmodemappings = !empty($viewmodemappings) ? (array) $viewmodemappings : [];
-      usort($viewmodemappings, ['\Drupal\format_strawberryfield\Form\ViewModeMappingSettingsForm', 'sortSettings']);
-
-      foreach ($viewmodemappings as $viewmodemapping) {
-        if (($viewmodemapping['active'] == TRUE) && in_array($viewmodemapping['jsontype'],$adotype)) {
-          $view_mode = $viewmodemapping['view_mode'];
-          break;
-        }
-      }
-    }
-    // To be determined if we want to be dependat on ds module or not
-    // if so we really want to act on this alter and not the general one.
-    /* \Drupal::moduleHandler()->alter('ds_switch_view_mode', $view_mode, $original_view_mode, $entity); */
+/**
+ * Implements hook_entity_view_mode_alter().
+ */
+function format_strawberryfield_entity_view_mode_alter(&$view_mode, Drupal\Core\Entity\EntityInterface $entity, $context) {
+  if ($entity->getEntityTypeId() != 'node') {
+    return;
   }
+  if ($view_mode != 'full') {
+    return;
+  }
+  $view_mode = \Drupal::service('format_strawberryfield.view_mode_resolver')->get($entity);
+
+  // To be determined if we want to be dependat on ds module or not
+  // if so we really want to act on this alter and not the general one.
+  /* \Drupal::moduleHandler()->alter('ds_switch_view_mode', $view_mode, $original_view_mode, $entity); */
+
 }
 
 /**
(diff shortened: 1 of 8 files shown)


pcambra:
Resolves #120

DiegoPino:
@pcambra I will review this during the weekend, thanks a lot!
We will discuss next week some performance concerns I have with this new implementation, probably better with numbers in place but may be just small things. Have a great weekend

pcambra:
@DiegoPino this is ready to review again :)

format_strawberryfield.module
```diff
@@ -36,15 +33,14 @@ function format_strawberryfield_page_attachments(array &$page) {
...
 
+/**
+ * Implements hook_form_FORM_ID_alter() for node_form.
+ */
 function format_strawberryfield_form_node_form_alter(&$form, FormStateInterface $form_state) {
-
   // Deal with Possible Addded Annotations.
-
   if ($form_state->get('hadAnnotations')) {
-    $node = $form_state->getFormObject()->getEntity();
```
DiegoPino, line 47: good. Can't remember why I left that there. Thanks for removing

format_strawberryfield.routing.yml
```diff
@@ -231,3 +231,18 @@ format_strawberryfield.deletetmp_webannotations:
     _format: 'json'
     _entity_access: 'node.update'
     _permission: 'add strawberryfield webannotation'
+
+# Display settings for each ADO
```
DiegoPino, line 235: Tiny thing. These are Display settings for bundles of type ADO, not for each ADO right?

src/Plugin/Block/DisplayInformationBlock.php
```diff
@@ -0,0 +1,158 @@
...
+                'view_mode_name' => $selected_view_mode['view_mode'],
+              ]
+            )->toString(),
+          ]
+        ),
+      ];
+      $build['info']['disclaimer'] = [
+        '#type' => 'html_tag',
+        '#tag' => 'div',
+        '#value' => $this->t('<strong>Warning</strong>: Changes on a View Mode will affect every ADO(Node) that is using it.'),
```
DiegoPino, line 113: Let's change this to "Info"

src/Plugin/Block/DisplayInformationBlock.php
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }
+
+    return $build;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
```
DiegoPino, line 139: lovely, all that uses caches is lovely

src/Plugin/Menu/LocalTask/ViewModeLocalTask.php
```diff
@@ -0,0 +1,65 @@
...
+      $plugin_id,
+      $plugin_definition,
+      $container->get('format_strawberryfield.view_mode_resolver')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getRouteParameters(RouteMatchInterface $route_match) {
```
DiegoPino, line 56: nice!

src/ViewModeResolver.php
```diff
@@ -0,0 +1,131 @@
...
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\strawberryfield\StrawberryfieldUtilityServiceInterface;
+
+/**
+ * View Mode resolver service.
+ *
+ * @ingroup format_strawberryfield
+ */
+class ViewModeResolver implements ViewModeResolverInterface {
+  use UseCacheBackendTrait;
```
DiegoPino, line 18: let's add a an extra line here before the trait , just for readability (sorry, I'm lacking coffee) 👴

src/ViewModeResolver.php
```diff
@@ -0,0 +1,131 @@
...
+    $ado_types = $this->getADOTypes($entity);
+    if (!empty($ado_types)) {
+      foreach ($this->getSortedMappings() as $mapping) {
+        if (($mapping['active'] == TRUE) && in_array($mapping['jsontype'], $ado_types)) {
+          return $mapping['view_mode'];
+        }
+      }
+    }
+
+    return 'default';
```
DiegoPino, line 74: Is default the right one? Where are the saved "view modes" stored when a user selects instead of the default (during edit/create) one?

src/ViewModeResolver.php
```diff
@@ -0,0 +1,131 @@
...
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   The content entity.
+   *
+   * @return array
+   *  Array of ado types.
+   *
+   * @throws \Exception
+   */
+  protected function getADOTypes(ContentEntityInterface $entity) {
+    $cache_id = 'format_strawberry:view_mode_adotypes:' . $entity->id();
```
DiegoPino, line 89: This cache should be per "JSON type" not per entity ID: on a 400,000 entity site the cache instead of helping will become a burden. If ADO or not is at the end a bundle type definition, and which view mode a JSON key "type" value one.

DiegoPino (changes requested):
Hi @pcambra thanks. I added mostly comments except for a larger one which requires intervention/code change (thus a request changes state here).

The cache can not be by entity->id(() If you go that route that cache will outlive its usefulness at generate a huge entry. The cache should be by "bundle identified as ADO" and JSON type found inside the flattened structure of the JSON. That way your cache will be Max of size of the different Json types x the view modes assigned x bundles bearing a SBF.

src/ViewModeResolver.php
pcambra: I'm afraid that comes from display suite, `$node->ds_switch->value`, and I'm not sure we want that dependency. Something like `\Drupal::service('entity_display.repository')->getViewDisplay($entity_type, $bundle, $view_mode);` will get us the default bundle display, but I'm not sure if it's enough.

src/ViewModeResolver.php
pcambra: @DiegoPino I don't see how this can be cached this way, as we're dependent on the entity. On this list we're caching the ADO types per entity as `bearsStrawberryfield` service still gets it as parameter. Seems like a catch 22, I can't cache before calling and I don't know the necessary information before caching.

DiegoPino: The egg/chicken dilemma (catch 22) is the reason we did not apply a cache previously for the json type itself when just having an entity. (the lack of a proper class of course at bundle level). Let me think about this a little deeper (like 24 hours, busy day preparing for tomorrow many workshop/zoom type of things) and give the whole thing a spin. I'm pretty sure there may be a pattern that is escaping us where performance and atomicity can play well. Thanks

pcambra:
@DiegoPino as an aside conversation, I've found that we're using the same code in https://github.com/esmero/ami/blob/main/src/AmiUtilityService.php#L253, we might want to do something about it

src/ViewModeResolver.php
pcambra: I've been thinking about this for a bit, as if the solution of caching it per entity is not acceptable, having a computed field wouldn't be either, I think that giving how aggressively Drupal caches everything, this would be fine but you obviously know the system better. Drupal performs so so with a lot of content without external caching and for memcache/redis, 400k simple entries like these are a breeze.

In any case, this is providing an additional cache layer that we don't currently have, so I would suggest three things here:

- Removing the cache layer.
- Adding a followup issue where we consider the field approach instead (this would replace the service)
- Having a place where we can actually test this; maybe an online version of a mid sized repository with something like new relic or the symfony one, where we can benchmark stuff?

DiegoPino:
@pcambra AMI has a lot of legacy code. That method is not in use right now and I have myself put the task of cleaning it up during this week.

That function is actually 1:1 to the view mode except its not about the view mode but about the actual type based on an UUID (AMI uses UUIDs to reference entities because IDs match also spreadsheet row numbers and it would be a mess, but in general UUIDs are better). The idea of that method (again, not used) is to fetch the JSON type key of a related (ismemberof, parof) Node and use semantics (not coded) to allow/now allow such relationships. But even if I have a full complex way of doing that via ontologies its really out of the scope of our work during 1.0.0. Ontologies (like "a bird can have wings and only 2 but not all things that have wings are birds") are cool and nice but (experience) will go over the head of most of our users, still working their way around JSON and Twig templates. Sorry long explanation and thanks for detecting that one.

src/ViewModeResolver.php
DiegoPino: > - Removing the cache layer.
Ok. I agree. Let's remove the cache layer for now. Caching which bundles are of ADO type may something easier we can approach (at least one check less) and that can go into the actual `::bearsStrawberryfield` method (another issue, not to be solved here)
> - Adding a followup issue where we consider the field approach instead (this would replace the service)

The field approach you mean for saving the view mode at the Node level? Instead of using the ds-> config? Yes. may even an option that people want to opt/in/out. We may keep the service and build a field check there if we go that route. Better to have more options than less.

> - Having a place where we can actually test this; maybe an online version of a mid sized repository with something like new relic or the symfony one, where we can benchmark stuff?
I agree. We can feed one of our repos a lot of data and even create a test batch ingest set (via the JSON API, a special costume module) so we can benchmark this. I once created 1,000,000 random objects in 2018 for this (when I decided on this architecture) but it was a brute force approach. Should be see this as a DevOps task with @alliomeria collaboration for the data?

src/ViewModeResolver.php
DiegoPino: Yeah, I had my coffee(s). Not even worth the change. Dismiss!

src/Plugin/Block/DisplayInformationBlock.php
DiegoPino: @pcambra just a few things pending here before I can merge. The Warning to Info is important so we do not scare the users!

src/ViewModeResolver.php
DiegoPino: I will mark the cache discussion as resolved for now until we have larger numbers and we can measure if cache v/s dynamically calculating this every time changes or not things. Thanks a lot of https://github.com/esmero/format_strawberryfield/pull/122/commits/bc84fe57d9fea3ca18f73f719159b824254bfc4d Which was needed.

DiegoPino (approved):
@pcambra approving this, but won't merge until the "Warning" v/s "info" message is changed. Thanks a lot, I know this was hard work and I appreciate the level of detail in your code!

pcambra:
@DiegoPino apologies, I'd swear I had changed that

pcambra:
@DiegoPino this should be ready to go I think

PR #129, opened 2021-02-22
ISSUE-123: Preliminary work to get a "Preview" feature on Metadata Display Entity edit forms

### LINKED ISSUE #123: Enable an ADO Entity Loader and a Preview functionality on the Metadata Display Entity Edit Form
# What is needed?

To make users writing and testing twig templates easier we want to add a few new features to the Metadata Display Entity Edit form
- An ADO autocomplete (we may need to check if we  already have the autocomplete handler, if not we may write one, simple task). This is to select any ADO object from the repository to load .. (continued)
- The SBF field Metadata/JSON into a second Text area, so users can see the sources and values present. This helps with Writing/testing twig templates
- A dual column layout, left the twig template (as it is now), right the loaded JSON to be used to preview/test the Template
- A "Preview button" that submits to a handler that does exactly what our Metadata Formatter does, render the Template using the JSON as context. Also passes our globals like the IIIF server.
- A tiny info area (fieldset) showing users which globals and their values are present
- Remove the redirect on "save" OR add a "save and continue" button,  open to both options, so every time a user saves its not redirected to main Entity List of the Metadata Display entity

@pcambra please let me know if you have any questions here. I tried to explain the use case as good as I could. The purpose is only one: help site builders and metadata professionals to build better twig templates by having data to see and test against.

@alliomeria anything else we could need to aid in the twig template creation?

this relates the more advanced issue #72 but that one is a separate thing.

pcambra:
This change includes:

- Preview feature on Metadata Display entity edit form, on a details that can be expanded, the "Show Preview" button will display an offcanvas with the HTML produced by the twig edits.
- On this particular edit form codemirror will now autosave on keyup as otherwise the ajax buttons would not reflect the changes correctly.
- CSS resets are removed from the offcanvas so the codemirror widget can be displayed correctly. See https://www.drupal.org/project/drupal/issues/2826722
- Minor tweaks and coding standard fixes

Fixes #123


src/Plugin/Field/FieldFormatter/StrawberryMetadataTwigFormatter.php
```diff
@@ -236,30 +207,29 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
       $json_error = json_last_error();
       if ($json_error != JSON_ERROR_NONE) {
-        $message= $this->t('We could had an issue decoding as JSON your metadata for node @id, field @field',
+        $message = $this->t('We could had an issue decoding as JSON your metadata for node @id, field @field',
           [
             '@id' => $nodeid,
             '@field' => $items->getName(),
           ]);
-        return $elements[$delta] = ['#markup' => $this->t('ERROR')];
+        return $elements[$delta] = ['#markup' => $message];
```
DiegoPino, line 229: wow. Gosh. good catch!

src/Form/MetadataDisplayForm.php
```diff
@@ -11,69 +19,205 @@
...
+    $entity = $form_state->getFormObject()->getEntity();
+
+    // Attach the library necessary for using the OpenOffCanvasDialogCommand and
+    // set the attachments for this Ajax response.
+    $form['#attached']['library'][] = 'core/drupal.dialog.off_canvas';
+    $form['#attached']['library'][] = 'codemirror_editor/editor';
+    $response->setAttachments($form['#attached']);
+
+    $input = $form_state->getUserInput();
+    $id = EntityAutocomplete::extractEntityIdFromAutocompleteInput($input['ado_context_preview']);
```
DiegoPino, line 167: good. we use this in webform sbf too.

src/Form/MetadataDisplayForm.php
```diff
@@ -11,69 +19,205 @@
...
+      '#value' => $entity->renderNative($context),
+      '#codemirror' => [
+        'lineNumbers' => FALSE,
+        'toolbar' => FALSE,
+        'readOnly' => TRUE,
+        'mode' => 'text/html',
+      ],
+    ];
+
+    $response->addCommand(new OpenOffCanvasDialogCommand(t('Preview'), $output, ['rezisable' => true, 'width' => '50%']));
```
DiegoPino, line 219: This is a typo:
Should be  'resizable' according to https://api.drupal.org/api/drupal/core%21modules%21outside_in%21src%21Ajax%21OpenOffCanvasDialogCommand.php/class/OpenOffCanvasDialogCommand/8.2.x

src/Plugin/Field/FieldFormatter/StrawberryMetadataTwigFormatter.php
```diff
@@ -289,53 +259,40 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
-
     }
+
     return $elements;
   }
 
   /**
    * {@inheritdoc}
    */
-  public function view(FieldItemListInterface $items, $langcode = NULL) {
```
DiegoPino, line 324: yep. basically a stub, good to remove

DiegoPino (changes requested):
@pcambra this looks pretty good 👏  I'm not finished reviewing and need to test now. I see there is good work here and effort and I value that. In the meantime I marked a tiny typo you can fix immediately while I go deeper. Thanks!

src/Form/MetadataDisplayForm.php
pcambra: I guess it is the default :)

DiegoPino: Since you have `rezisable` instead of `resizable`, you can either remove it (and let the default take over) or fix the typo. Up to you

pcambra: Pushed the fix

format_strawberryfield.routing.yml
```diff
@@ -45,7 +45,7 @@ format_strawberryfield.metadatadisplay_settings:
   path: 'admin/structure/metadatadisplay_entity_settings'
   defaults:
     _form: '\Drupal\format_strawberryfield\Form\MetadataDisplaySettingsForm'
-    _title: 'Metadata Display Settings'
+    _title: 'Metadata display settings'
```
DiegoPino, line 48: I just noticed this. Is it some Drupal convention to use names in lower case?

src/Form/MetadataDisplayForm.php
```diff
@@ -11,69 +19,205 @@
...
+    /** @var \Drupal\format_strawberryfield\MetadataDisplayInterface $entity */
+    $entity = $form_state->getFormObject()->getEntity();
+
+    // Attach the library necessary for using the OpenOffCanvasDialogCommand and
+    // set the attachments for this Ajax response.
+    $form['#attached']['library'][] = 'core/drupal.dialog.off_canvas';
+    $form['#attached']['library'][] = 'codemirror_editor/editor';
+    $response->setAttachments($form['#attached']);
+
+    $input = $form_state->getUserInput();
```
DiegoPino, line 166: Question, is the `ado_context_preview` not in the $form-state as value? Wonder if getting data from the User Input, not validated may bring issues? (testing that now)

src/Form/MetadataDisplayForm.php
```diff
@@ -11,69 +19,205 @@
...
+        }
+        else {
+          $context['data'][$offset] = $jsondata;
+        }
+      }
+    }
+    // Ensure we're using the twig from user's input instead of the entity's
+    // default.
+    $entity->set('twig', $input['twig'][0]);
+    $output = [
```
DiegoPino, line 207: Cool.  I feel this pull is almost ready, only one missing need right now and I'm totally open to suggestions. Ping @alliomeria too.
As explained in the ISSUE description we need to show the user the RAW JSON too (like as as in the twig fiddle example) because it is the `$context` they need to be able to pick values and do the actual Twig template writing. This fits (render array) here but we could also maybe use a horizontal/vertical tab to allow both to be displayed using the same height and selectable? Or some other UI/UX idea? Access to the RAW JSON as one writes/codes templates on the right side is quite important (in terms of interface). If having both there is too complex I feel we could even move the Preview itself down and give the RAW values priority. Thoughts you both?

alliomeria: Hello @pcambra & @DiegoPino! Having the selected ADO's raw json available for quick review while writing Twig templates would be very helpful. A section (maybe collapsable?) above the new Preview area on the right hand side could be a comfortable setup.

src/Form/MetadataDisplayForm.php
DiegoPino: Thanks!

format_strawberryfield.routing.yml
pcambra: I noticed the "core" ones are that way, just my OCD kicking in

src/Form/MetadataDisplayForm.php
pcambra: That's because on ajax requests, multisteps etc, the values are not guaranteed to be there. As the user is already a privileged one I don't think this is going to bring much issues. It is a good point that we're not validating the twig itself so we might want to do that? not sure if it's worth it.

src/Form/MetadataDisplayForm.php
pcambra: I am a bit confused about this, you're saying that we need somewhere to add the json from the SBF? I will try to implement what you're suggesting, let's see if I got it right

pcambra:
@DiegoPino @alliomeria is this what you were thinking about?


https://user-images.githubusercontent.com/104912/108839848-de1f2900-75d5-11eb-8a0b-719a17c366f6.mov

Wondering if we could benefit of the codemirror search plugin https://codemirror.net/demo/search.html

alliomeria:
@pcambra This looks great to me! I can understand that it might seem like too much screen-clutter with the 3 separate areas, but having the full json readily available when creating & editing the metadata display templates will be extremely helpful.

format_strawberryfield.routing.yml
DiegoPino: Will make an habit of that. Thanks!

src/Form/MetadataDisplayForm.php
DiegoPino: We are validating the twig at the entity level. Here: https://github.com/esmero/format_strawberryfield/blob/main/src/Entity/MetadataDisplayEntity.php#L278

and also catching render exceptions (which is an implicit Twig template validation) here https://github.com/esmero/format_strawberryfield/pull/129/files/b4ce99ee2bd874025b193b4581dc611e18bea926#diff-376335f1b2fb78915a415da3636afd7ce6a2927d608c5bda585db4cbce30d91fR124

I can concord with not a strict validation if $form_state is unreliable because of the Code Mirror and it seems like ::extractEntityIdFromAutocompleteInput by doing regular expression may be enough. I will roll with getUserInput() as you have it now. Thanks

src/Form/MetadataDisplayForm.php
DiegoPino: Yes Pedro. It is in the ISSUE description and its super important for the editing experience. Without that you need to have another window with the RAW data visible because Twig templating will vary if a Key is there, if the value is single or plural. What I do not know is if 3 textures is the best OR some type of collapsible or even a replacement (a swap between JSON and the Preview). Mostly because once you see the Preview, you are Kinda Done with it until you press "preview button again". I will leave this decision to @alliomeria since she is the one that uses this most and can speak for the metadata community. Thanks!

DiegoPino:
@pcambra forget my comment about where to put the JSON. That place/collapsible fields works well for me if it works for @alliomeria. Thanks for addressing that!

src/Form/MetadataDisplayForm.php
DiegoPino: Your new code addresses this 🥇 . Thanks Pedro! Marking this as resolved!!

src/Form/MetadataDisplayForm.php
```diff
@@ -201,10 +201,29 @@ public static function ajaxPreview($form, FormStateInterface $form_state) {
...
+    $output = [];
+    $output['json'] = [
+      '#type' => 'details',
+      '#title' => t('JSON Data'),
+      '#open' => false,
+    ];
+    $output['json']['data'] = [
+      '#type' => 'codemirror',
+      '#rows' => 60,
+      '#value' => json_encode($context['data'], JSON_PRETTY_PRINT),
```
DiegoPino, line 214: This is actually quite cool. The fact that you are showing the whole $context and not only the SBF JSON. Just noticed it and makes me =) Thanks!

DiegoPino (approved):
Seems ready to go. Will do a last sanity check/human test and will merge later today. Great work @pcambra

DiegoPino:
Oh, don't forget during this period all goes against RC2. Once RC2 is out, we move to RC3 or 1.0.0 if we feel like super stars. I will update that now!

PR #130, opened 2021-02-25
ISSUE-125: Full screen view cut in half w/ OpenSeadragon viewer

### LINKED ISSUE #125: Full screen view cut in half w/ OpenSeadragon viewer
When I view assets in full screen mode on a large monitor, the viewer cuts to be half the screen rather than full screen. Appears to happen only with the OpenSeadragon viewer.

pcambra:
This PR removes the container class from the iiif div so the width is always 100%. 
Also some minor coding style fixes

Fixes #125

### DIFF
--- a/src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
@@ -233,8 +233,6 @@ class StrawberryMediaFormatter extends StrawberryBaseFormatter {
                 // yet.
                 // @see https://github.com/esmero/format_strawberry/issues/1
 
-                //@ TODO recheck cache tags here, since we are not really using
-                // the file itself.
                 $iiifpublicinfojson = $iiifhelper->getPublicInfoJson($iiifidentifier);
 
                 $groupid = 'iiif-' . $items->getName() . '-' . $nodeuuid . '-' . $delta . '-media';
@@ -253,6 +251,11 @@ class StrawberryMediaFormatter extends StrawberryBaseFormatter {
                     'data-iiif-group' => $grouped ? $groupid : $uniqueid,
                     'data-iiif-thumbnails' => $thumbnails,
                     'style' => "width:{$max_width_css}; height:{$max_height}px",
+                  ],
+                  //@ TODO recheck cache tags here, since we are not really using
+                  // the file itself.
+                  '#cache' => [
+                    'tags' => $file->getCacheTags(),
                   ],
                 ];
                 if (isset($item->_attributes)) {
@@ -291,7 +294,8 @@ class StrawberryMediaFormatter extends StrawberryBaseFormatter {
                   // This also never runs if cached. So after deletion we better
                   // call the controller!
                   if (!empty($jsondata['ap:annotationCollection']) && is_array($jsondata['ap:annotationCollection'])) {
-                    WebAnnotationController::primeKeyStore($items[$delta]);
+                    $keystoreid = $elements[$delta]['media' . $i]['#attached']['drupalSettings']['format_strawberryfield']['openseadragon'][$uniqueid]['keystoreid'];
+                    WebAnnotationController::primeKeyStore($items[$delta], $keystoreid);
                   }
                 }
                 if ($this->currentUser) {
--- a/src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Created by PhpStorm.
  * User: dpino
@@ -16,6 +17,7 @@ use Drupal\strawberryfield\Tools\StrawberryfieldJsonHelper;
 use Drupal\Core\StreamWrapper\StreamWrapperManager;
 use Drupal\format_strawberryfield\Controller\WebAnnotationController;
 use Drupal\Core\Url;
+
 /**
  * Simplistic Strawberry Field formatter.
  *
@@ -32,6 +34,7 @@ use Drupal\Core\Url;
  * )
  */
 class StrawberryMediaFormatter extends StrawberryBaseFormatter {
+
   /**
    * {@inheritdoc}
    */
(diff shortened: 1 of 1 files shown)


src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
```diff
@@ -259,53 +258,57 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
                   $elements[$delta]['media' . $i]['#attached']['drupalSettings']['format_strawberryfield']['openseadragon'][$uniqueid]['keystoreid'] = WebAnnotationController::getTempStoreKeyName($fieldname, $delta, $nodeuuid);
                   $elements[$delta]['media' . $i]['#attached']['drupalSettings']['format_strawberryfield']['openseadragon'][$uniqueid]['webannotations'] = (boolean) $webannotations;
                   $elements[$delta]['media' . $i]['#attached']['drupalSettings']['format_strawberryfield']['openseadragon'][$uniqueid]['webannotations_tool'] = $webannotations_tool ? $webannotations_tool : 'rect';
-                  // This also never runs if cached. So after deletion we better call the controller!
+                  // This also never runs if cached. So after deletion we better
+                  // call the controller!
                   if (!empty($jsondata['ap:annotationCollection']) && is_array($jsondata['ap:annotationCollection'])) {
-                    $keystoreid = $elements[$delta]['media' . $i]['#attached']['drupalSettings']['format_strawberryfield']['openseadragon'][$uniqueid]['keystoreid'];
```
DiegoPino, line 288: This here seems to be a bug/of mine
WebAnnotationController::primeKeyStore() should be primed with the $keystoreid. Issue was this whole thing is cached and gets actually never call again after a cache clear. So wondering if it can be removed totally OR, better not and we need to prime it, at least once to avoid problems when the cache expires (e.g because of Node edit).

Can you please change to to WebAnnotationController::primeKeyStore($keystoreid) ?

src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
```diff
@@ -151,30 +155,27 @@ public function settingsSummary() {
...
   }
 
-
   /**
    * {@inheritdoc}
    */
   public function viewElements(FieldItemListInterface $items, $langcode) {
     $elements = [];
     $max_width = $this->getSetting('max_width');
-    $max_width_css = empty($max_width) || $max_width == 0 ? '100%' : $max_width .'px';
```
DiegoPino, line 165: yep. what a shame!  sloppy iterative code. Thanks!

src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
```diff
@@ -215,41 +216,39 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
                 if ($iiifidentifier == NULL || empty($iiifidentifier)) {
                   continue;
                 }
-                // ImageToolKit use the $file->getFileUri(), we don't want that yet
+                // ImageToolKit use the $file->getFileUri(), we don't want that
+                // yet.
                 // @see https://github.com/esmero/format_strawberry/issues/1
 
-                //@ TODO recheck cache tags here, since we are not really using the file itself.
-                $filecachetags = $file->getCacheTags();
```
DiegoPino, line 240: Wonder if we should add the file cache tags instead of removing this? (Given its and not Solved `@TODO`) or do you think it is not needed?

src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
```diff
@@ -215,41 +216,39 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+                $groupid = 'iiif-' . $items->getName() . '-' . $nodeuuid . '-' . $delta . '-media';
+                $uniqueid = $groupid . $i;
 
-                $elements[$delta]['media'.$i] = [
+                $elements[$delta]['media' . $i] = [
                   '#type' => 'container',
                   '#default_value' => $uniqueid,
                   '#attributes' => [
                     'id' => $uniqueid,
-                    'class' => ['strawberry-media-item','field-iiif','container'],
```
DiegoPino, line 252: And this the actual fix! Good. Good.

DiegoPino (commented):
Awesome @pcambra. singe css fix but with the DCS it touches a lot of code so took me a bit longer to review.
There is one line you removed that needs to go back but also needs to be fixed (and issue on my part) and tested (by clearing caches and starting new annotations, the changing user, etc) the code  way this works is also good to explore for you. Thanks a lot

Also, about the file cache. We normally do not remove `@TODO` except if we are over them or we get them fixed or they are not relevant.  In this case I do not know enough to be sure if the file cache should be taken in account here or not. If you feel its not needed all good and we are over this, but if not, either leave the comment there or if you feel so add them a cache tag(s) for the actual render array.

src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
pcambra: Aw sorry about that, we're not doing anything with the tags so maybe some more work is needed for this?

DiegoPino: No problem. Let's keep the //@TODO for now and remove 240. We can work on that in another issue. Thanks

DiegoPino: Hey. I will merge this. We can deal with the @TODO another day. I have done the cache stuff on the WARC formatter so we can copy from there. Thanks!

DiegoPino:
Still need to wait for tomorrow to merge because of the non initialized (my fault) storage. Let's get that done and we move from there. I will merge the Twig editor in the meantime!

src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
pcambra: I've added the cache tags to the element delta, is that what would be intended?

src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
pcambra: Done

src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
```diff
@@ -291,7 +294,8 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
                   // This also never runs if cached. So after deletion we better
                   // call the controller!
                   if (!empty($jsondata['ap:annotationCollection']) && is_array($jsondata['ap:annotationCollection'])) {
-                    WebAnnotationController::primeKeyStore($items[$delta]);
+                    $keystoreid = $elements[$delta]['media' . $i]['#attached']['drupalSettings']['format_strawberryfield']['openseadragon'][$uniqueid]['keystoreid'];
+                    WebAnnotationController::primeKeyStore($keystoreid);
```
DiegoPino, line 298: Hi Pedro, sorry again. I explained myself wrong here: It is both, look at the primeKeyStore signature. 
```PHP
/**
   * Primes the Web Annotation KeyStore with saved values.
   *
   * @param \Drupal\strawberryfield\Plugin\Field\FieldType\StrawberryFieldItem $itemfield
   * @param null $keystoreid
   */
  public static function primeKeyStore(StrawberryFieldItem $itemfield,  $keystoreid = NULL) {
    if ($keystoreid == NULL && strlen(trim($keystoreid)) == 0) {
      return NULL;
    }
    $jsondata = $itemfield->provideDecoded(TRUE);
    $tempstore = \Drupal::service('tempstore.private')->get(
      'webannotation'
    );
    if (!empty($jsondata['ap:annotationCollection']) && is_array($jsondata['ap:annotationCollection'])) {
      $tempstore->set($keystoreid, $jsondata['ap:annotationCollection']);
      return $jsondata['ap:annotationCollection'];
    }
    else {
      $tempstore->set($keystoreid, []);
      return [];
    }
  }
``` this won't work that way. You can look at how we call this in other parts.

The issue I had was that I was passing only the $items[$delta] when we already know where we want the keystore to be primed. But also this needs some little testing if you could. Thanks

pcambra: Ah I see that now, not sure why PHPStorm didn't call me out when I did that

src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php
DiegoPino: Almost! Sorry again (what a day!) its not $elements[$delta] its $items[$delta], $elements[$delta] its an render array, $items[$delta] the actual field with JSON data.

pcambra: Oops sorry about that!

DiegoPino:
And merging. Thanks @pcambra !

Issue #121, opened 2021-02-04
Excessive drift while using Pannellum in VR


powerman5500:
Using an Oculus Quest 2, when navigating the Pannellum panoramas the images "drift" after the controller button has been released.  The problem occurs both on instances of Pannellum within Archipelago as well as the demo instance on pannellum.org, in fullscreen mode as well as in the embedded window. 

The "friction" setting referenced in Pannellum documentation at https://pannellum.org/documentation/reference/ may be able to reduce the drifting but I'm not certain.

Issue #109, opened 2020-12-04
Make replay web formatter act on WACZ instead of WARC if present


DiegoPino:
@giancarlobi this is a place holder for the formatter need.

Right now its acting on WARC files, now that we have automatic WARC to WACZ transformations we need to adapt code to react to the fact that there could be 2 files with the same content but one WAY faster and with indexed content!

DiegoPino:
@pcambra @alliomeria @dmer @giancarlobi this is a question for you all (please respond, not now but eventually)

Right now our Formatters are acting quite simplistically on the media (files) present. Certain formatters/viewers can handle Images, others PDFs, others use Twig template generated manifests. This one in particular WARC files.

We want to be finer in what we fetch/how we fetch the media for the ones that are not driven by IIIF manifests and access files directly.

If you look at our `as:document` or `as:image` JSON structure you will notice a key that is empty (me forward thinking but I have 2 hands only and a wandering brain). The key is  "tags"


```JSON
  "as:image": {
        "urn:uuid:ab6f3b46-9a74-4089-bf0c-4304312074ab": {
            "url": "s3:\/\/d1a\/image-image-view-ab6f3b46-9a74-4089-bf0c-4304312074ab-ab6f3b46-9a74-4089-bf0c-4304312074ab.tiff",
            "name": "view.tiff",
            "tags": [],
            "type": "Image",
            "dr:fid": 64,
            "dr:for": "images",
            "dr:uuid": "ab6f3b46-9a74-4089-bf0c-4304312074ab",
            "checksum": "d1ae6b04d4d12027dc9247b46071b974",
```

The idea here was to allow users, during ingest (specially webform) to pass tags (e.g "display", "preservation", "hide"), hopefully controlled but may also be free form, or dictated by a vocab, to uploaded files. There are two issues in terms of UI for this. Any of these structures are generated after the actual ingest. So having a special element / JSON KEY that can hold these tags/values will, at least for a little while during processing, imply a different place for that storage in the JSON key structure and imply a challenge too for AMI/Batch ingests (UI wise, the actual code is not complex).

Also, because as:image is a helper structure, made to be "destroyed" and rebuild  (magically) on demand if the user/you/I decide or need to.

[...]

PR #131, opened 2021-02-26
ISSUE-109: Allow JMESPaths to define which media is used. WebReplay

DiegoPino:
# What is this?

See #109 

A new base class for Formatters that need/want to extract Media from the JSON in a more fine grained way. This allows right now two extra JMESPaths to be defined and also 2 fallback options.

This also adds non related data that we found today. IIIF limits on Pannellum panorama Formatter/JS. Now we detect if the Device is Mobile or Not and also calculate (gosh) estimated memory usage based on the IIIF JSON provide size arrays. That way we avoid Java HEAP overflows with + 32K images when asking for more that 50%. you can always without any issues as for MAX size since those are served directly without any preprocessing/scaling from Cantaloupe. I also added + friction to help with #121

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 
 src/Plugin/Field/FieldType/IIIFlinkItem.php
 .idea/format_strawberryfield.iml
+.idea/codeStyles/Project.xml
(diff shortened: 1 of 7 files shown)


DiegoPino:
Ok, merging this because I kinda need it know. RC2 is not released yet so if I made a mistake here please blame me for ever, I can take it

PR #139, opened 2021-03-03
ISSUE-138: Add file tags cache to strawberry_paged_formatter (books/documents)

### LINKED ISSUE #138: Add file tags cache to strawberry_paged_formatter (books/documents)
When rendering the Paged formatter, adding the file tags to the cache render array element will allow us to invalidate the file cache an keep all the metadata intact.

This issue has a fine pairing 🍷 with https://github.com/esmero/strawberry_runners/issues/25

pcambra:
- Adds file cache tags to the render array of the paged formatter
- Adds new method "getFile" that could be reused on the base formatter
- Minor cleanup on baseformatter, pdf and video

Resolves #138

### DIFF
--- a/src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Created by PhpStorm.
  * User: dpino
@@ -8,6 +9,7 @@
 
 namespace Drupal\format_strawberryfield\Plugin\Field\FieldFormatter;
 
+use Drupal\Core\Cache\Cache;
 use Drupal\Core\Field\FieldItemListInterface;
 use Drupal\Core\Field\FieldItemInterface;
 use Drupal\Core\Form\FormStateInterface;
@@ -17,6 +19,7 @@ use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\strawberryfield\StrawberryfieldUtilityServiceInterface;
+use Drupal\strawberryfield\Tools\Ocfl\OcflHelper;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Drupal\Core\Template\TwigEnvironment;
 use Drupal\Core\Field\FieldDefinitionInterface;
@@ -40,13 +43,6 @@ use Drupal\strawberryfield\Tools\StrawberryfieldJsonHelper;
 class StrawberryPagedFormatter extends StrawberryBaseFormatter implements ContainerFactoryPluginInterface {
 
   /**
-   * The current user.
-   *
-   * @var \Drupal\Core\Session\AccountInterface
-   */
-  protected $currentUser;
-
-  /**
    * The entity manager.
    *
    * @var \Drupal\Core\Entity\EntityTypeManagerInterface
@@ -54,6 +50,8 @@ class StrawberryPagedFormatter extends StrawberryBaseFormatter implements Contai
   protected $entityTypeManager;
 
   /**
+   * The Twig environment.
+   *
    * @var \Drupal\Core\Template\TwigEnvironment
    */
   protected $twig;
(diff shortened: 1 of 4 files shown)


src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -125,7 +122,7 @@ public static function defaultSettings() {
     return parent::defaultSettings() + [
       'iiif_group' => TRUE,
       'mediasource' => 'json_key',
-      'json_key_source' => 'as:image',
+      'json_key_source' => 'as:document',
```
pcambra, line 125: This is the part I'm a bit more nervous about, not sure if the as:image was a leftover or it is intentional as I've seen the other formatters rely heavily on the key source to get the file from the metadata in json

DiegoPino:
@pcambra this is good and still there is a lot here to review and discuss. I will probably need a few days (sorry) to get around all the use cases, performance considerations and to test. I will move on comment at the time and we should (again) have a call about this to explain a few cases where this won't work out of the box (sadly)
Caching/performance is always a beast I know and the code is good. It is not about the particular code. I will start adding comment in a few. Thanks!

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -515,10 +493,41 @@ public function processElementforManifestURL($delta, array $jsondata, FieldItemI
...
+   * @param string $type
+   *   Relevant type for the formatter.
+   *
+   * @return \Drupal\file\FileInterface|bool
+   *   File if found, FALSE otherwise,
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  protected function getFile(array $mediaitem, string $type) {
```
pcambra, line 516: I think this needs to use the field_file_drop, opinions? @DiegoPino

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
DiegoPino: Sorry for the late response here: Bookreader (paged) can work on both, images and Documents, so please leave image as default. We ship with examples of this and we have a few in play.archipelago.nyc too

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -292,60 +293,45 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
       if (empty($value)) {
         continue;
       }
-      /* @var array $jsondata */
       $jsondata = json_decode($item->value, TRUE);
       // @TODO use future flatversion precomputed at field level as a property
       $json_error = json_last_error();
       if ($json_error != JSON_ERROR_NONE) {
-        return $elements[$delta] = ['#markup' => $this->t('ERROR')];
```
DiegoPino, line 300: good

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
DiegoPino: Hi, `field_file_drop` is very useful but makes picking a Document v/s an Image v/s a 3D Model too complex (means checking the mime types and deciding there after iterating over each one. Its better used when we do not care what type of file we are dealing with but more like "files" in general.

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
DiegoPino: Ok, I retract my comment. This is a left over (I can't see track of all my own code apparently). We do not need that value anymore. Issue is originally we have 4 strategies. (still in the comments), as:image, as:document, metadata display and external URL IIIF manifest.

Since this one came https://github.com/esmero/format_strawberryfield/pull/29 in (and that one, please look at it) was more like a rescue operation after a too large pull got merged (🤕 ) with bugs and I had to fix it ASAP we have now only 2 operations.

Which also means... since we are only using the "twig template" or the external URL... gosh... we only need to use the actual per NODE cache here, since all is driven by what the JSON provides. If a file gets changed/removed its also about the JSON itself that gets updated. 

Conclusion:
1.- YES, good catch and good you were nervous: we do not need/nor use that Key anymore. Means the form needs to be gone, the setting and even the schema? (schema changes are a mess.. so good with making that change too? I guess it will also imply a hook update to remove the setting)
2.- All your other changes are great but we can't use File Cache here at all. I mean, it is the same as the JSON cache which is the same as the NODE cache so its redundant (sorry)

and
protected function getFile(array $mediaitem, string $type)  (which is a great function) should go to the parent class so we can use it where its really needed, e.g the Media Formatter - Open Seadragon (which uses directly media) or to this new class, which will be used for anything where access to media is dynamic (now only WARC) and https://github.com/esmero/format_strawberryfield/blob/1.0.0-RC2/src/Plugin/Field/FieldFormatter/StrawberryDirectJsonFormatter.php . Up to you

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
pcambra: I know nothing! 👯

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
pcambra: Should we just close this PR and open separate issues for the components that are actually useful then?

DiegoPino: Ok, it may be the best. Let's gather and recover energy and take over smaller chunks of this. Sorry! But it was a nice discussions and we learned so much!

DiegoPino:
@pcambra per you request,
https://github.com/esmero/format_strawberryfield/pull/139#discussion_r596284111

let's close this one in the meantime and revisit later with more smaller issues focusing on improving each viewer. I really appreciate all the work behind of this one, truly.

Issue #145, opened 2021-03-22
Add twig extension for load entity by label


patdunlavey:
# What?
When importing data via AMI, where users want to reference an existing entity, such as a taxonomy term, user, or node, we do not want to require that they know and use entity ids in the data that they are importing. Users expect to be able to put the name of the entity in the spreadsheet, and have it recognized as the term or node that they expect. 

I propose that a twig extension `\Drupal\format_strawberryfield\TwigExtension::load_entities_by_label` be added. It would take as arguments the label, entity type and bundle identifier as required arguments, and return an array of entity objects of type `\Drupal\Core\Entity\EntityInterface` whose labels match the provided label.

Sample usage in the ami twig template:
```
{% if data.term_name %}
    {% set terms = load_entities_by_label(data.term_name, 'taxonomy_term', 'my_vocabulary') %}
    {% if terms[0] %}
    "my_term":  {{ terms[0].tid|json_encode|raw }},
    {% endif %}
{% endif %}
```

@DiegoPino does this make sense to you?

DiegoPino:
@patdunlavey interesting. is data.term_name a single label? We may want to add access control to the implementation. To make sure all our Twig extensions do not collide with other modules we will want to prefix them with `sbf_` or an `ap_` (short of archipelago).
What happens if that load matches like 1000? (e.g people with Page 1), maybe adding a limit argument? with a small default?

patdunlavey:
I had already intended to include a limit argument, with a default of 1.

Here's what I have written (not tested at all!!!!): https://github.com/Born-Digital-US/format_strawberryfield/blob/ISSUE-145/src/TwigExtension.php#L26-L105

DiegoPino:
@patdunlavey I see you already have your own pull! So will comment here. I would suggest going for 
https://api.drupal.org/api/drupal/core%21lib%21Drupal.php/function/Drupal%3A%3AentityQuery/9.2.x
instead of direct access to the Database: 

Check https://api.drupal.org/api/drupal/core%21lib%21Drupal.php/function/Drupal%3A%3AentityQuery/9.2.x

I would also add a MAX to the output. On render time fetching 10,000 seems bad idea.
And finally (sorry, this is a thing I'm planning on every other new extension) re-check that in NO case your Extension throws an exception, and if so, catch it, and return NULL. and Log it.



@pcambra does my suggestion seem appropriate?

patdunlavey:
[...]

PR #147, opened 2021-03-24
WIP: Issue 145 -- Add twig extension for load entity ids by entity label

patdunlavey:
# What?

This is a simplified version of the original feature request: https://github.com/esmero/format_strawberryfield/issues/145. Rather than load entity data, this twig extension just loads entity ids of entities whose labels match the provided name.

# To test
Put something like the following example into your twig template, with tests for each of the supported entity types: node, taxonomy_term, user, group. 

Example: Given a field in the twig data with a name of a taxonomy term, get the first taxonomy term ids from all terms with that label in a particular vocabulary:
```
{% if data.term_name %}
    {% set terms = sbf_entity_ids_by_label(data.term_name, 'taxonomy_term', 'my_vocabulary') %}
    {% if terms[0] %}
    "my_term":  {{ terms[0]|json_encode|raw }},
    {% endif %}
{% endif %}
```

### DIFF
--- a/src/TwigExtension.php
+++ b/src/TwigExtension.php
@@ -1,6 +1,7 @@
 <?php
 namespace Drupal\format_strawberryfield;
 use Twig\TwigTest;
+use Drupal\twig_tweak\TwigExtension as TwigTweakExtension;
 
 /**
  * Class TwigExtension.
@@ -34,7 +35,11 @@ class TwigExtension extends \Twig_Extension {
    *   The entity type.
    * @param  string  $bundle_identifier
    *   The entity bundle (may be empty)
-   * @param  numeric  $limit
+   * @param  string  $view_mode
+   *   The view mode for the render array that should be returned for each entity.
+   * @param  bool  $check_access
+   *
+   * @param  int  $limit
    *   Restrict to number of results.
    * @param  null  $langcode
    *   (optional) For which language the entity should be rendered, defaults to
@@ -42,13 +47,16 @@ class TwigExtension extends \Twig_Extension {
    *
    * @return null|[Drupal\Core\Entity\EntityInterface]
    *   An array of entity objects or NULL if no entity with label exists.
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
-  public function load_entities_by_label(string $label, string $entity_type, string $bundle_identifier = '', $limit = 1, $langcode = NULL): ?array {
+  public function load_entities_by_label(string $label, string $entity_type, string $bundle_identifier = '', $view_mode = 'default', $check_access = TRUE, $limit = 1, $langcode = NULL): ?array {
     $label = \Drupal::database()->escapeLike($label);
     switch($entity_type) {
       case 'node':
-        $query = \Drupal::entityQuery('node')
-          ->condition('title', $label, 'LIKE')
+        $query = \Drupal::entityTypeManager()->getStorage('node')->getQuery();
+        $query->condition('title', $label, 'LIKE')
+          ->accessCheck($check_access)
           ->range(0,$limit);
         if($bundle_identifier) {
           $query->condition('type', $bundle_identifier);
@@ -56,8 +64,9 @@ class TwigExtension extends \Twig_Extension {
         $ids = $query->execute();
         break;
       case 'taxonomy_term':
-        $query = \Drupal::entityQuery('taxonomy_term')
-          ->condition('name', $label, 'LIKE')
+        $query = \Drupal::entityTypeManager()->getStorage('taxonomy_term')->getQuery();
+        $query->condition('name', $label, 'LIKE')
+          ->accessCheck($check_access)
           ->range(0,$limit);
         if($bundle_identifier) {
           $query->condition('vid', $bundle_identifier);
@@ -65,8 +74,9 @@ class TwigExtension extends \Twig_Extension {
         $ids = $query->execute();
         break;
       case 'group':
-        $query = \Drupal::entityQuery('group')
-          ->condition('label', $label, 'LIKE')
+        $query = \Drupal::entityTypeManager()->getStorage('group')->getQuery();
+        $query->condition('label', $label, 'LIKE')
+          ->accessCheck($check_access)
           ->range(0,$limit);
         if($bundle_identifier) {
           $query->condition('type', $bundle_identifier);
@@ -74,26 +84,21 @@ class TwigExtension extends \Twig_Extension {
         $ids = $query->execute();
         break;
       case 'user':
-        $query = \Drupal::entityQuery('user')
-          ->condition('name', $label, 'LIKE')
+        $query = \Drupal::entityTypeManager()->getStorage('user')->getQuery();
+        $query->condition('name', $label, 'LIKE')
+          ->accessCheck($check_access)
           ->range(0,$limit);
         $ids = $query->execute();
         break;
     }
 
     if(!empty($ids)) {
-      $entities = $this->getEntityTypeManager()->getStorage($entity_type)->loadMultiple($ids);
-      if (empty($entities)) {
-        return NULL;
+      $entities = [];
+      $twig_tweak = new TwigTweakExtension();
+      foreach($ids as $id) {
+        $entities[$id] = $twig_tweak->drupalEntity($entity_type, $id, $view_mode, $langcode);
       }
-
-      // Get the entities in the specified context language.
-      $entityRepository = $this->getEntityRepository();
-      $translated_entities = [];
-      foreach($entities as $entity) {
-        $translated_entities[$entity->id()] = $entityRepository->getTranslationFromContext($entity, $langcode);
-      }
-      return $translated_entities;
+      return $entities;
     }
 
     return NULL;
--- a/src/TwigExtension.php
+++ b/src/TwigExtension.php
@@ -44,44 +44,40 @@ class TwigExtension extends \Twig_Extension {
    *   An array of entity objects or NULL if no entity with label exists.
    */
   public function load_entities_by_label(string $label, string $entity_type, string $bundle_identifier = '', $limit = 1, $langcode = NULL): ?array {
-    $database = \Drupal::database();
+    $label = \Drupal::database()->escapeLike($label);
     switch($entity_type) {
       case 'node':
-        $query = $database->select('node_field_data', 'n')
-          ->addField('n', 'nid', 'id')
-          ->condition('n.title', $database->escapeLike($label), 'LIKE')
+        $query = \Drupal::entityQuery('node')
+          ->condition('title', $label, 'LIKE')
           ->range(0,$limit);
         if($bundle_identifier) {
-          $query->condition('n.type', $bundle_identifier);
+          $query->condition('type', $bundle_identifier);
         }
-        $ids = $query->execute()->fetchCol();
+        $ids = $query->execute();
         break;
       case 'taxonomy_term':
-        $query = $database->select('taxonomy_term_field_data', 't')
-          ->addField('t', 'tid', 'id')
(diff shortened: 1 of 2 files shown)


src/TwigExtension.php
```diff
@@ -22,4 +24,94 @@ public function is_instanceof($value, string $type): bool
...
+    }
+
+    catch(\Exception $exception) {
+      $responseMessage = $exception->getMessage();
+      $message = $this->t('@exception_type thrown in @file:@line while querying for @entity_type entity ids matching "@label". Message: @response',
+        [
+          '@exception_type' => get_class($exception),
+          '@file' => $exception->getFile(),
+          '@line' => $exception->getLine(),
+          '@entity_type' => $entity_type,
```
DiegoPino, line 103: bundle is missing

src/TwigExtension.php
```diff
@@ -22,4 +24,94 @@ public function is_instanceof($value, string $type): bool
...
+          $query->condition('name', $label, 'LIKE')
+            ->accessCheck($check_access)
+            ->range(0,$limit);
+          $ids = $query->execute();
+          break;
+      }
+    }
+
+    catch(\Exception $exception) {
+      $responseMessage = $exception->getMessage();
```
DiegoPino, line 97: This intermediate var is not needed, can be assigned directly to '@response' =>

src/TwigExtension.php
```diff
@@ -22,4 +24,94 @@ public function is_instanceof($value, string $type): bool
...
+   *
+   * @return null|array
+   *   An array of render arrays for the entities found, or NULL if the entity does not exist.
+   */
+  public function sbf_entity_ids_by_label(string $label, string $entity_type, string $bundle_identifier = '', $check_access = TRUE, $limit = 1, $langcode = NULL): ?array {
+    $label = \Drupal::database()->escapeLike($label);
+    $limit = min($limit, 100);
+    /** @var \Drupal\Core\Entity\Query\QueryInterface $query */
+    try {
+      switch($entity_type) {
```
DiegoPino, line 55: This is more a suggestion in terms of reducing code (not against repeating code when it adds to legibility, so take/leave)

You allow a set of Entities here, 'node', etc. Maybe checking with an in_array() first to return immediately if not present.
Then the $query = \Drupal::entityTypeManager()->getStorage($valid_type)->getQuery(); would be a single call.
And these are also common

```
->accessCheck(TRUE) // hope so!
->range(0,$limit); 
```

independently of the type
```
And same for  $ids = $query->execute();
```

your only, per type particularities are which field contains the "title" and "how" the $bundle_identifier needs to be processed. That can stay in your switch for sure.

That would reduce a lot of code

src/TwigExtension.php
```diff
@@ -1,5 +1,7 @@
 <?php
 namespace Drupal\format_strawberryfield;
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Cache\CacheableMetadata;
```
DiegoPino, line 4: Where is this used?

src/TwigExtension.php
```diff
@@ -22,4 +24,94 @@ public function is_instanceof($value, string $type): bool
       || (\function_exists($func = 'is_'.$type) && $func($value))
       || $value instanceof $type;
   }
-}
\ No newline at end of file
+
+  /**
+   * Returns entity ids of entities with matching title/name/label.
+   *
+   * Supported entity types are node, taxonomy_term, group, and user.
```
DiegoPino, line 31: This may go under "entity_type" as a comment

src/TwigExtension.php
```diff
@@ -22,4 +24,94 @@ public function is_instanceof($value, string $type): bool
...
+            $query->condition('type', $bundle_identifier);
+          }
+          $ids = $query->execute();
+          break;
+        case 'user':
+          $query = \Drupal::entityTypeManager()->getStorage('user')->getQuery();
+          $query->condition('name', $label, 'LIKE')
+            ->accessCheck($check_access)
+            ->range(0,$limit);
+          $ids = $query->execute();
```
DiegoPino, line 91: Just for safety I would also add 
`->addTag('distinct');`
to make sure you do not get duplicates. May be just me being paranoid!

DiegoPino (changes requested):
@patdunlavey good! I really like how this is looking! I have a few request, Most are my general Code concerns without testing and some recommendations based on a quick exploration of the methods used. Happy to get a "no, no, mr" and discuss further! HUGS! 
PS: a final DCS could also help, I saw a single issue, but a fast pass is recommended =)

src/TwigExtension.php
```diff
@@ -36,80 +45,56 @@ public function is_instanceof($value, string $type): bool
...
+      'group' => ['label', 'type'],
+      'user' => ['name', NULL],
+    ];
+    $label_field = $fields[$entity_type][0] ?? NULL;
+    if ($label_field) {
+      $bundle_field = $fields[$entity_type][1] ?? NULL;
+      $limit = min($limit, 100);
+      /** @var \Drupal\Core\Entity\Query\QueryInterface $query */
+      try {
+        $query = \Drupal::entityTypeManager()->getStorage('node')->getQuery();
```
DiegoPino, line 72: @patdunlavey this is looking so good. Thanks for addressing my code review, I really appreciate it. You may be working on this already,  but just in case because I will forget it mysql later:

`getStorage('node')` should be after your refactor `getStorage($entity_type')`

Let me know when you want me to do some testing. Again, thanks and good work :Clarkson

src/TwigExtension.php
```diff
@@ -22,4 +24,94 @@ public function is_instanceof($value, string $type): bool
...
+   * @param  int  $limit
+   *   Restrict to number of results. Capped at no more than 100.
+   * @param  null  $langcode
+   *   (optional) For which language the entity should be rendered, defaults to
+   *   the current content language.
+   *
+   * @return null|array
+   *   An array of render arrays for the entities found, or NULL if the entity does not exist.
+   */
+  public function sbf_entity_ids_by_label(string $label, string $entity_type, string $bundle_identifier = '', $check_access = TRUE, $limit = 1, $langcode = NULL): ?array {
```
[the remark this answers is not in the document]
patdunlavey: Thanks for your thorough review!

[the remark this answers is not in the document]
patdunlavey: This is much changed now. I removed the access check logic since apparently entityQuery include access check by default, and I'm convinced that there is no need to override that.

src/TwigExtension.php
```diff
@@ -22,4 +24,94 @@ public function is_instanceof($value, string $type): bool
...
+    $label = \Drupal::database()->escapeLike($label);
```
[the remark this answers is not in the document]
patdunlavey: I'd used "LIKE" thinking that it was necessary to get a case-insensitive match. It turns out that the '=' operator is case-insensitive by default for text.

src/TwigExtension.php
```diff
@@ -22,4 +24,94 @@ public function is_instanceof($value, string $type): bool
...
+    $limit = min($limit, 100);
+    /** @var \Drupal\Core\Entity\Query\QueryInterface $query */
+    try {
+      switch($entity_type) {
+        case 'node':
+          $query = \Drupal::entityTypeManager()->getStorage('node')->getQuery();
+          $query->condition('title', $label, 'LIKE')
+            ->accessCheck($check_access)
+            ->range(0,$limit);
```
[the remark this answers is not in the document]
patdunlavey: Check.

src/TwigExtension.php
patdunlavey: I thought that it's unlikely that we would see an error from the entityQuery due to something with the bundle, so I left it out. Besides, not all entity types have bundles or their equivalent. E.g users.

src/TwigExtension.php
patdunlavey: Check.

src/TwigExtension.php
patdunlavey: I totally rewrote this section along the lines you suggested.

src/TwigExtension.php
patdunlavey: Not used. Removed.

src/TwigExtension.php
patdunlavey: Done.

src/TwigExtension.php
patdunlavey: I don't see how it's possible to get duplicates from a single table?

src/TwigExtension.php
patdunlavey: Oops! How'd that sneak back in there!

patdunlavey:
@DiegoPino I tried to respond to all your comments -- all excellent!! This remains WIP since I _still_ have not gotten to test it out!

src/TwigExtension.php
patdunlavey: If you're inclined to try testing, I think it's ready for that @DiegoPino .

src/TwigExtension.php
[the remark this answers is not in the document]
DiegoPino: Great! Resolved

src/TwigExtension.php
DiegoPino: Thanks so much. Good coding

src/TwigExtension.php
DiegoPino: True (but code is always an unknown). I agree with you, saw some core example with it but not even an issue here! I concede and concord 👍

DiegoPino:
@patdunlavey this is all very good. I will find some time tomorrow while having my coffee and toasted twigs to test. Looks super good and hard to break but will make sure I'm thorough. Thanks so much 🏆

src/TwigExtension.php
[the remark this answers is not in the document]
DiegoPino: Thanks to you for good disposition on this. I appreciate highly the patience and addressing all my concerns.

DiegoPino:
@patdunlavey hi, good Sunday for you.

Did some testing (around) and found only tiny issues. I have a pull I can make against your branch if that works or do you prefer comments and you correct them? 

I have also question. For the label, are you targeting exact matches or "Contains"? This right now works only for Exact ones (which is Ok for me since will yield less "randomness" but want to be sure that is what you want =)

Big hugs!  Pull coming if you are Ok with that, or code review comments, all after  my coffee Numero 5.

patdunlavey:
I'm happy to go with either approach - whichever you prefer. I am very curious as to what you had to do to solve the setError problem.

Regarding using "contains", I don't think that would be useful, since the purpose is to be as unambiguous as possible when using an entity label to find the entity id. I can't think of a real world use case where a fuzzy match would be useful, and there's the danger that we would match a term that we don't want ahead of the term that we do want (we're not doing a solr search, so there's no "score" result).

DiegoPino:
@patdunlavey ok. will prepare a pull and let you know. Just tiny things that came up during testing. FYI: the setError follows the way other compound elements do it, by going deep into the compound structure. But I also found another thing in the element (Duplicated Titles) that I want to fix at the same time (sorry) so that pull will come up late night or tomorrow morning. (sorry again)

patdunlavey:
Hi @DiegoPino I *think* your last comment refers to a [different PR](https://github.com/esmero/webform_strawberryfield/pull/104). 

I finally got to test this myself (ha!), and it is working for me as I expect. Namely, given a string in the twig source data, I was able to use it to load a taxonomy term ID like this example:
```twig
      {% set ids = sbf_entity_ids_by_label(data.obj_print_item_parts__ip_print_price_bundle, 'taxonomy_term', 'price_bundle') %}
      {% if ids | length > 0 %}
        "ip_print_price_bundle":  {{ ids | first | json_encode | raw }},
      {% endif %}
```

DiegoPino:
@patdunlavey sorry, yes! Can't find my way around anymore. I did do a few days ago a pull against yours. Mostly DCS. https://github.com/Born-Digital-US/format_strawberryfield/pull/1
In case you feel any of that is needed or you can take/leave but makes sense to you there.

DiegoPino (approved):
Good work!

PR #148, opened 2021-03-29
ISSUE-136: Adds Validation, error messages and Rendered Preview to Metadata Display Preview

DiegoPino:
# What?

Continuing on #123, this pull adds contextual errors, clears accumulative errors/messages while previewing, adds a Native Render option with basic (very basic) validation, Twig Template Syntax contextual errors and messages to aid the User in editing/previewing Metadata Display Entities. Also adds mimetype modes to CodeMirror (JSON, JSON-LD, XML)

WIP still, I may do better tomorrow (theming for warnings) but seems to work as expected so far.

e.g

Twig error:

image

Rendered with HTML preview

image

Wrong Entity Selected

image

Wrong JSON
image


DiegoPino:
@alliomeria @giancarlobi I will merge this now. If there are other issues I can work on those. Will never be perfect because UI interaction may always be a little be unexpected but it works fair enough. Thanks!

Issue #149, opened 2021-04-07
Permit sbf_entity_ids_by_label twig extension to handle both entity name and id as input


patdunlavey:
The purpose of the `sbf_entity_ids_by_label` twig extension is to permit csv data to be more human-readable. However, sometimes people just export the data using values stored in the json (entity ids) and never need to see the human-readable names of those entities. In this case, we should not lose the ids when it is re-imported. 

So I propose that the `sbf_entity_ids_by_label` twig function add a boolean argument "fail_to_id" which, if true (the default), and if the label value fails to match the label of an existing entity, and if it is numeric, then gets tested to see if it is the entity id of an existing entity. If so, then the twig function simply returns the value as the found entity id.

I've written code for this, and will create a pull request momentarily.

PR #150, opened 2021-04-07
ISSUE-149: Added 'fail_to_id' argument to \Drupal\format_strawberryfield\TwigExtension::entityIdsByLabel

patdunlavey:
See https://github.com/esmero/format_strawberryfield/issues/149

# What:
See ISSUE-149 for rationale.

This adds an argument to \Drupal\format_strawberryfield\TwigExtension::entityIdsByLabel, `$fail_to_id`, to tell the function that if
- $fail_to_id is true (default)
- the $label value fails to match an existing entity
- the $label value is numeric
...then check to see if the value is an entity id. If so, then just return the value.

The biggest code changes are around exception handling, since we now have two places that could trigger an exception. So I moved the exception handler to a separate method.

@DiegoPino, I decided to use  `\Drupal::entityTypeManager()->getStorage($entity_type)->load($label)` rather than another EntityQuery, mainly because I'm lazy and don't want to have to hard code entity id field names. I suspect there's a better way to do that which doesn't involve actually loading the entity, though I didn't see an obvious one looking at EntityStorageBase.

### DIFF
--- a/src/TwigExtension.php
+++ b/src/TwigExtension.php
@@ -42,15 +42,17 @@ class TwigExtension extends \Twig_Extension {
   /**
    * Returns entity ids of entities with matching title/name/label.
    *
-   * @param string $label
+   * @param  string  $label
    *   The entity label that we're looking for
-   * @param string $entity_type
+   * @param  string  $entity_type
    *   The entity type.
    *   Supported entity types are node, taxonomy_term, group, and user.
-   * @param string $bundle_identifier
+   * @param  string  $bundle_identifier
    *   The entity bundle (may be empty)
-   * @param int $limit
+   * @param  int  $limit
    *   Restrict to number of results. Capped at no more than 100.
+   * @param  bool  $fail_to_id
+   *   Should we check if an unmatched label value is actually an entity id?
    *
    * @return null|array
    *   An array of Entity IDs for the entities found, or NULL if no match.
@@ -59,7 +61,8 @@ class TwigExtension extends \Twig_Extension {
     string $label,
     string $entity_type,
     string $bundle_identifier = '',
-    $limit = 1
+    $limit = 1,
+    $fail_to_id = TRUE
   ): ?array {
     $fields = [
       'node' => ['title', 'type'],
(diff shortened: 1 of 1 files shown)


DiegoPino:
@patdunlavey this may be just me but maybe you do not need this (you may say yes and we go ahead, no worries)
Let me explain: 
- In a label based operation you are Getting the ID for a label. And you are passing already the Entity Type and the Bundle of any, and your extension will return the numeric ID.

So, what if for the case your input is a number (or many numbers)

You try in the twig template https://github.com/esmero/format_strawberryfield/blob/bd84b20798106fc4f5761cedc27dc1a3124b10f3/src/TwigExtension.php#L26-L30
```Twig
{% for ids_or_label in data.my_ids_or_labels %}
  {% if ids_or_label is instanceof("scalar") %} 
  // do the logic you need for numbers. Maybe load them via bamboo? Or you can go for "int" to be strict? Maybe just pass them through?
  {% else %}
 // Call the label resolver 
  {% endif %}
{% endfor %}
````

Just a silly idea.

patdunlavey:
Thanks @DiegoPino . I know it's an edge case, but entity labels can be numers. So I think we need to actually look to see if there is an entity with that label, even if the label is a number. Only if that fails, then look to see if it's an entity id. A further consideration: if we assume that numeric entity labels are entity ids, and don't actually check if it's a valid entity id, then we could pass through an invalid entity id.

DiegoPino:
@patdunlavey let me think about this. Because if your Labels can be a number, then how do you know what was intended by the user? There can be a label "22" and also a term "22"... Maybe, for the sake of consistency you could use entity paths for non labels entity:taxonomy:2 ? 

The issue about invalid entity id's could need a different extension. One that uses a list of Ids in an entity query and returns only back the ones that match.  I kinda feel it is better to have (for your edge cases) separate functions instead of functions with too many arguments and logic. Let's talk about this in Slack tomorrow morning and figure out the Pros and Cons of this, happy to get this reviewed and merged but I want to be sure it covers at least the "22" label v/s "22" id thing

DiegoPino:
@patdunlavey do you have any comments on this? Just wonder if you want this into RC2 and if my reply added just confusion. Let me know! Happy we can work this out

DiegoPino:
@patdunlavey this is stale. Do you mind closing it? Or maybe you went another route? Thanks!

patdunlavey:
Yes. Stale. Closing. Thanks!!

PR #154, opened 2021-05-05
1.0.0 rc2 textselect

giancarlobi:
@DiegoPino This for be able to use text select. I moved ocr check in formatter to be able to enable/disable text selection in js and avoid ajax sync deprecation. Please check and suggest anything you think is better to change. thanks friend


giancarlobi:
@DiegoPino Not clear to me why also pannellum related changes appears here, my branch comes from RC2 aligned to esmero RC2 ... if this is a problem I can delete this, sorry again

giancarlobi:
We need laso this https://github.com/esmero/strawberryfield/pull/176 to make text selection working

js/iiif-iabookreader_strawberry.js
```diff
@@ -32,25 +35,19 @@
...
+                    textSelection: {
+                      enabled: has_ocr,
+                      singlePageDjvuXmlUrl: '/do/' + node_uuid + '/flavorsearch/all/ocr/djvuxml/{{pageIndex}}',
+                    },
+                  },
                   padding: 11,
                 };
                 var br = new BookReader(options);
                 br.init();
-                // Check if Book has or not OCR using Ajax callback
```
DiegoPino, line 39: @giancarlobi this needs to come back. We are using Ajax to identify if there is OCR

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -371,6 +371,7 @@ public function processElementforMetadatadisplays($delta, array $jsondata, Field
     $element = [];
     $entity = NULL;
     $nodeuuid = $item->getEntity()->uuid();
+    $has_ocr = $this->searchEnabled($item);
```
DiegoPino, line 374: We are not using this anymore. If we use this flag then it will be cached. That means once OCR is available it won't show

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -432,6 +433,7 @@ public function processElementforMetadatadisplays($delta, array $jsondata, Field
         $element['media']['#attributes']['data-iiif-infojson'] = '';
         $element['media']['#attached']['drupalSettings']['format_strawberryfield']['iabookreader'][$htmlid] = [
           'nodeuuid' => $nodeuuid,
+          'has_ocr' => $has_ocr,
```
DiegoPino, line 436: same here. We need to remove this

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -462,6 +464,7 @@ public function processElementforManifestURL($delta, array $jsondata, FieldItemI
     $element = [];
     $entity = NULL;
     $nodeuuid = $item->getEntity()->uuid();
+    $has_ocr = $this->searchEnabled($item);
```
DiegoPino, line 467: And here too

DiegoPino (commented):
@giancarlobi yes, this is bringing old code back. We need only your changes in https://github.com/esmero/format_strawberryfield/pull/154/files#diff-f1334b4b90fc2cdbc05134d69d024cdffc2311e6d6cb30eaf25e1d4fa718bfa3 and the Bookreader JS. And there is conflict. If you want I can clean up the conflict manually.

giancarlobi:
> @giancarlobi yes, this is bringing old code back. We need only your changes in https://github.com/esmero/format_strawberryfield/pull/154/files#diff-f1334b4b90fc2cdbc05134d69d024cdffc2311e6d6cb30eaf25e1d4fa718bfa3 and the Bookreader JS. And there is conflict. If you want I can clean up the conflict manually.

I saw that we need 4.40.3 instead of 4.21.0 and that is the conflict, if you can please manually resolve. Thanks

js/iiif-iabookreader_strawberry.js
giancarlobi: @DiegoPino The issue is that we need to know if ocr exists also to enable/disable text selection some lines above, for that reason I changed to this code, if not, we have to make a sync call before init to enable/disable text selection or any other solution you consider ok

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
giancarlobi: Ok, I understand it, so to enable/disable text selection if no ocr we have to make a sync ajax call before setting options in js, right?

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
giancarlobi: ok

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
giancarlobi: ok

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
DiegoPino: Yes, we do that in https://github.com/esmero/format_strawberryfield/blob/66bbd57e9d90982e46bb925d49640693aba27dc9/js/iiif-iabookreader_strawberry.js#L40-L50
So for your use case we need to do it before so you can pass the argument to the `br`.

Means that function can return has_ocr (which can start being FALSE always).  Does that work?

giancarlobi: @DiegoPino I made that before, it works but I received a console deprecation because of ajax sync in main thread so I switched to this last solution, bad, so I can pull my first code whit deprecation, what do you think?

DiegoPino: You are right, async: true needs to be the right one now for $.ajax({  (but it is the default in recent Jquery, what version do you have?).

We can force it, but if so, we need to put the full br initialization inside the Ajax response
Or one may trigger before the Book reader is already there. We can try/test. Does that make sense?
Who loves JS? not me!

DiegoPino: Something like

$.ajax({     url : "example.com",    async : true /* default is true */ }).done(function(response){     // Start BR})

DiegoPino:
Closing since your code is in 1.3.0 now. Thanks so much for all your work on Archipelago @giancarlobi

Issue #177, opened 2021-12-09
ADO with no media skips embargo context


aksm:
# What?

Thanks to @giancarlobi testing Drupal 9.3.0, a bug was discovered where ADOs with no media can cause something like the following error:

```
A notice: Notice: Undefined variable: embargo_context in Drupal\format_strawberryfield\Plugin\Field\FieldFormatter\StrawberryMediaFormatter->viewElements() (line 244 of /var/www/archipelago/web/modules/contrib/format_strawberryfield/src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php) #0 /var/www/archipelago/web/core/includes/bootstrap.inc(346): _d
12:53
Then error: TypeError: Argument 3 passed to Drupal\Core\Cache\Cache::mergeContexts() must be of the type array, null given, called in /var/www/archipelago/web/modules/contrib/format_strawberryfield/src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php on line 244 in Drupal\Core\Cache\Cache::mergeContexts() (line 29 of /var/www/archipelago/web/core/lib/Drupal/Core/Cache/Cache.php) #0 /var/www/archipelago/web/modules/contrib/format_strawberryfield/src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php(244): Drupal\Core\Cache\Cache::mergeContexts() #1 /var/www/archipelago/web/core/lib/Drupal/Core/Field/FormatterBase.php(89): Drupal\format_strawberryfield\Plugin\Field\FieldFormatter\StrawberryMediaFormatter->viewElements() #2 /var/www/archipelago/web/core/lib/Drupal/Core/Entity/Entity/EntityViewDisplay.php(263): Drupal\Core\Field\FormatterBase->view()
```

@DiegoPino has already identified the likely culprit here:

https://github.com/esmero/format_strawberryfield/blob/1d2e51d2b3d0ef07f5f6fbf5d12499e7f62b7895/src/Plugin/Field/FieldFormatter/StrawberryMediaFormatter.php#L207

DiegoPino:
Fixed via https://github.com/esmero/format_strawberryfield/pull/178

PR #178, opened 2021-12-09
ISSUE-177: Initialize embargo cache arrays sooner

DiegoPino:
See #177. This really applies to any Formatter and only fails if there is no media at all to show. This fixes it but @aksm @alliomeria @giancarlobi I'm not totally happy still with the refactor for formatters. I might push another (different issue) before 1.0.0

@giancarlobi can you please test tomorrow?

Thx


giancarlobi:
@DiegoPino it works perfect, no more errors. Thanks best bug solver friend!

giancarlobi (approved):

Issue #185, opened 2022-04-04
Enable bookreader image size other than MAX; append GET requests


patdunlavey:
Work in [this PR](https://github.com/esmero/format_strawberryfield/pull/181) failed to include an ability for the bookreader to use a image width other than the full image width. It also does not pass on any request arguments.

See discussion starting [here](https://github.com/esmero/format_strawberryfield/pull/181#issuecomment-1086267338).

patdunlavey:
Closing this issue the associated PR since they have become stale.

PR #186, opened 2022-04-04
ISSUE-185--bookreader-image-width-args --- 

patdunlavey:
Responding to [ISSUE-185](https://github.com/esmero/format_strawberryfield/issues/185):

utilize canvas width in and append image request arguments to image url for getImagesListApi3().

Is this something like what you had in mind, @DiegoPino ?

### DIFF
--- a/js/plugin.iiif-iabookreader_strawberry.js
+++ b/js/plugin.iiif-iabookreader_strawberry.js
@@ -311,7 +311,7 @@ BookReader.prototype.parseSequence = function (sequenceId) {
               }
               if(annotation.body.id) {
                 imageObj.imageGetArgument = getURLArgument(annotation.body.id);
-                imageObj.imageUrl = annotation.body.id.replace(/\/full\/full\/0\/default.jpg/, '/full/'+ imageObj.canvasWidth + ',/0/default.jpg') + imageObj.imageGetArgument;
+                imageObj.imageUrl = annotation.body.id.replace(/\/full\/full\/0\/default.jpg/, '/full/'+ imageObj.canvasWidth + ',/0/default.jpg');
               }
               else {
                 imageObj.imageUrl = "";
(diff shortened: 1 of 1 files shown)


DiegoPino:
@patdunlavey did you test this with PDFs and other stuff?

patdunlavey:
Thanks to your inquiry, I found that it was *not* working for PDFs because I had appended the urlArguments when they were already there. I didn't notice it on non-PDF objects because they don't have a query appended to the url. I believe that is now fixed with my latest commit.

The canvas width defaults to 0 if it's not provided, which I think is how it works in 2.1:
https://github.com/esmero/format_strawberryfield/blob/0e8949d442ddc6bb789357ee298a9dcd4cd853b4/js/plugin.iiif-iabookreader_strawberry.js#L301-L302
Are you saying it should skip pages whose width is not set (0)?

Regarding your question about trying it with other stuff, I added an audio object to my test collection that I have configured to display using the "IIIF Presentation API 3 Creative Works Series Manifest" and Bookreader. All the image objects display, but no "page" is created for the audio object in the bookviewer. But I'm pretty sure that's because that metadatadisplay template needs work. Do you think this bookreader js may need further work to display other object types, assuming we adapt the metadatadisplay templates to provide all the necessary/possible canvas data? If so, can we make a new issue for that?

DiegoPino:
@patdunlavey we need to go back to this discussion. There are a few things that might not needed and also the "showing" other media here is an issue of IABook Reader. Could we extend it? Sure. We can take the Mirador approach and show an HTML5 Video/Audio Player as output. Could we maintain it? How many of us too keep this alive between IABook reader release? That is my main concern.

patdunlavey:
Hi @DiegoPino as best I can make out, this PR is no longer applicable. Shall we close it?

DiegoPino:
@patdunlavey yes please. If this is an issue/future problem we tackle it again. Thanks and sorry for the late reply.

patdunlavey:
Will do. Thanks!!

Issue #87, opened 2020-08-17
CSL-citations as a twig extension & a field formatter: a tale of two converging paths


DiegoPino:
# What?
So you want to allow people to show-render-expose CSL citations. We all do. Where is the most obvious place to put them?

- As a simple Twig extension so people can directly render (cool) from their metadata displays well formed citations
- Using the same logic but a bit more elaborated example, via a full formatter.

Gist: 
Make a dependency
- https://github.com/seboettg/citeproc-php
- Write the extension and document it
- Write a formatter that uses a list of twig templates (full customizable!) and has a select/box/checkbox and some JS so copy to clipboard works, changing works etc.
Test, add the twig templates to out list of auto deployables and a config

Make IR dance. Done

@alliomeria and @giancarlobi this is IR work. I feel citations can be important.

DiegoPino:
Hey, i'm diego, i would love to follow up here.

DiegoPino:
Resolved

PR #192, opened 2022-04-26
ISSUE-87: CSL-citations as a twig extension & a field formatter: a tale of two converging paths

aksm:
Beginning work on #87. Still a long way to go. Currently some of the styles (list of examples below) cause the first letter of each word in a title to be stripped:

chicago-author-date
chicago-author-date-16th-edition
university-of-york-mla

It's possible that the following log notice is involved:
```
Notice: iconv(): Wrong charset, conversion from `ISO-8859-1' to `UTF-8//IGNORE' is not allowed in Symfony\Polyfill\Mbstring\Mbstring::mb_convert_case() (line 285 of /var/www/html/vendor/symfony/polyfill-mbstring/Mbstring.php).
```


aksm:
@DiegoPino I think I located bug involving words in titles being stripped of their first letters.

The following ternary evaluates to true, but then the true condition doesn't return the concatenated string (I'm guessing because of that Mbstring notice that occurs in the logs (mentioned above):

```php
return in_array($encoding, self::ISO_ENCODINGS) ?
    Mbstring::mb_strtoupper($firstChar, $encoding).$then : $firstChar.$then;
```

The above is from here:

https://github.com/seboettg/citeproc-php/blob/afdfbfd616b2ed3dd0f0e45feda72d0b007e9c8e/src/Util/StringHelper.php#L142-L144

Still looking into it. Will follow up as soon as I figure out more.

DiegoPino:
@aksm the problem is this here 
https://stackoverflow.com/questions/63746822/why-does-iconv-returns-empty-string-in-php7-4-fpm-alpine-docker

aksm:
@DiegoPino Hmm, I saw that and skipped, hoping it wasn't a devops/dependency issue. Do we go with the accepted answer?

DiegoPino:
Probably good to test with the accepted answer. Try running the command on your own local php container (without rebuilding, just the command live) and reload the container. See if that changes things. IF so, then I will rebuild both 7.4 and 8.0 to include an updated iconv library

src/Plugin/Field/FieldFormatter/StrawberryCitationFormatter.php
```diff
@@ -0,0 +1,588 @@
...
+    //$citation_locales_file = '/var/www/html/vendor/citation-style-language/locales/locales.json';
+    //$citation_locales_file_contents = file_get_contents($citation_locales_file);
+    //$citation_locales = json_decode($citation_locales_file_contents, true);
+    //$locale_options = array();
+    //foreach($citation_locales['language-names'] as $key => $value) {
+    //  $locale_string = $value[1];
+    //  $locale_options[$key] = $this->t($locale_string);
+    //}
+    // There's a better way to get this directory
+    $citation_style_directory = '/var/www/html/vendor/citation-style-language/styles-distribution';
```
DiegoPino, line 146: `citation-style-language` library needs to be a composer dependency and yes, that is not a good way of getting that folder

DiegoPino: I will fix my own comment here. If we go composer we will need a custom script to git clone that repository. Or we can ship this as a DevOps operation and have a configuration (global one for archipelago) to state where this is downloaded/present

DiegoPino: Ok, this is already solved in the libraries example

{
    "name": "vendor-name/program-name",
    "repositories": [
        {
            "type": "package",
            "package": {
                "name": "citation-style-language/locales",
                "version":"1.0.0",
                "source": {
                    "type": "git",
                    "url": "https://github.com/citation-style-language/locales.git",
                    "reference": "master"
                }
            }
        },
        {
            "type": "package",
            "package": {
                "name": "citation-style-language/styles-distribution",
                "version":"1.0.0",
                "source": {
                    "type": "git",
                    "url": "https://github.com/citation-style-language/styles-distribution.git",
                    "reference": "master"
                }
            }
        }
    ],
    "require": {
        "citation-style-language/locales":"@dev",
        "citation-style-language/styles-distribution":"@dev",
        "seboettg/citeproc-php": "^2"
    }
}, so this needs to go in our composer, not just the "seboettg/citeproc-php": "^2"

aksm:
Hi @DiegoPino, partly out of curiosity about whether it was possible and partly out of expediency to get things working quickly, I inlined all the HTML and Javascript for the Select element.

Will gladly refactor if you think this is a bad idea, but it will probably take me some time to understand (sorry, I'm a slow learner!) all the parts that you showed me and get it functioning.

aksm:
@DiegoPino Do I have this right for generating a unique ID for the element, or am I missing something?

DiegoPino:
> @DiegoPino Do I have this right for generating a unique ID for the element, or am I missing something?

You mean referencing an existing ID as here `var theid = '#' + $(this).attr("id");` ?

aksm:
> > @DiegoPino Do I have this right for generating a unique ID for the element, or am I missing something?
> 
> You mean referencing an existing ID as here `var theid = '#' + $(this).attr("id");` ?

More on the other side, i.e. generating a unique ID that won't conflict if there are multiple blocks:

```php
      $uniqueid =
        'bibliography-' . $items->getName(
        ) . '-' . $nodeuuid . '-' . $delta;
```

src/Plugin/Field/FieldFormatter/StrawberryCitationFormatter.php
```diff
@@ -10,6 +10,7 @@
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Template\TwigEnvironment;
 use Drupal\format_strawberryfield\EmbargoResolverInterface;
+use Drupal\pathauto\MessengerInterface;
```
DiegoPino, line 13: Hi, this interface is not the right one for the Messenger. Check other plugins or inspect $this->messenger()

src/Plugin/Field/FieldFormatter/StrawberryCitationFormatter.php
```diff
@@ -133,20 +134,24 @@ public function settingsForm(array $form, FormStateInterface $form_state) {
       $entity = reset($entities);
     }
     $csl_root = DRUPAL_ROOT . '/libraries/citation-style-language';
-    $csl_exists = $this->fileSystem->prepareDirectory($csl_root);
-    $citation_style_directory = $csl_root . '/styles-distribution';
-    // Get the list of style files.
-    $style_list = $this->fileSystem->scanDirectory($citation_style_directory, '/\.(csl)$/i', ['recurse' => FALSE, 'key' => 'name']);
-    # Generate a list of select options and push in the styles.
+    $csl_exists = is_dir($csl_root);
```
DiegoPino, line 137: Good

src/Plugin/Field/FieldFormatter/StrawberryCitationFormatter.php
```diff
@@ -133,20 +134,24 @@ public function settingsForm(array $form, FormStateInterface $form_state) {
...
     $style_options = array();
-    foreach($style_list as $style) {
-      $style_name = $style->name;
-      $style_xml = simplexml_load_file($style->uri);
-      $style_title = $style_xml->info->title->__toString();
-      $style_options[$style_name] = $this->t($style_title);
+    if (!$csl_exists) {
+      $this->messenger()->addWarning('Please run "drush archipelago-download-citeproc-dependencies" before using this formatter.', 'warning');
+    } else {
```
DiegoPino, line 141: DCS (Drupal coding standards), the `else` needs to go in a next line

src/Plugin/Field/FieldFormatter/StrawberryCitationFormatter.php
```diff
@@ -133,20 +134,24 @@ public function settingsForm(array $form, FormStateInterface $form_state) {
...
+      $style_list = $this->fileSystem->scanDirectory($citation_style_directory, '/\.(csl)$/i', ['recurse' => FALSE, 'key' => 'name']);
+      # Generate a list of select options and push in the styles.
+      foreach($style_list as $style) {
+        $style_name = $style->name;
+        $style_xml = simplexml_load_file($style->uri);
+        $style_title = $style_xml->info->title->__toString();
+        $style_options[$style_name] = $this->t($style_title);
+      }
+      // Alphabetize them.
+      asort($style_options);
```
DiegoPino, line 153: Good!

src/Plugin/Field/FieldFormatter/StrawberryCitationFormatter.php
```diff
@@ -169,13 +174,15 @@ public function settingsForm(array $form, FormStateInterface $form_state) {
...
         '#options' => $style_options,
+        '#disabled' => !$csl_exists,
         '#default_value' => $this->getSetting('citationstyle'),
       ],
       'localekey' => [
         '#type' => 'textfield',
         '#title' => $this->t('Provide a metadata key to use as the locale (language) for citations.'),
         '#default_value' => $this->getSetting('localekey'),
         '#required' => FALSE,
+        '#disabled' => !$csl_exists,
```
DiegoPino, line 185: Very good

src/TwigExtension.php
```diff
@@ -274,6 +274,20 @@ public function bibliography(array $value, string $locale, array $styles = []):
...
+        'id' => 'bibliography' . 'test-unique-id',
+        'class' => ['bibliography'],
+      ],
+      '#attached' => [
+        'library' => 'format_strawberryfield/citations_strawberry'],
+    ];
+    $render_bibliography['bibliography'] = [
+      '#markup' => \Drupal\Core\Render\Markup::create($bibliography),
+    ];
+    $rendered_bibliography = \Drupal::service('renderer')->render($render_bibliography);
```
DiegoPino, line 290: @aksm this is great!!. Hint. Try returning it without the -> render and check it out. Twig should be able to render Render arrays directly. If that does not work out let me know!

aksm: @DiegoPino Tried that at one point and got the `Return value must be of type string` error. It looks like that only works with functions, not filters? Am I missing something?

aksm:
@DiegoPino I feel this is complete for the most part so will start on the next issue and come back with fresh eyes to review for mistakes/additions. Thanks!

Issue #196, opened 2022-05-06
Add a Copy from UI JS library to Citations/Metadata Formatters


DiegoPino:
@aksm this is going to be a follow up ticket to #192 

Many times we might need to have a tiny button on a side/surrounding an HTML element to allow the user (end user) to copy into its copy-pasta buffer easily data.

This can be used in the citation formatter but also on the Object Description (any Twig based, Custom Metadata Formatter) to copy the URL of a IIIF manifest, a persistent URL, etc.

The idea would be to make a small JS drupal behavior that attaches to a CSS class and uses a `data` attribute to get the to-be-copied value.

Browsers implement this differently. So other than the JS library, both CSL and the Custom Metadata Formatter would attach this library and we would need to add documentation for the former formatter so people know how to expose the HTML that allow the JS to attach itself to the HTML element.

Let me know if you have any doubts.

Thanks

DiegoPino:
Resolved

PR #200, opened 2022-05-27
ISSUE-196: Add a Copy from UI JS library to Citations/Metadata Formatters

aksm:
Starting work on #196


DiegoPino:
@aksm because this brings the same pulls as the CSL one, I need to review that one first and merge, so might take me until Sunday. So sorry, but will do, promised

DiegoPino (approved):
Good. Need to test but that is what a release is for!

Issue #203, opened 2022-06-03
Adds OpenCV functionality to Annotorious


DiegoPino:
# What?

Face detection + contour detection (two methods) via a OpenCV JS we worker + UI/UX. Extra cool!

DiegoPino:
Solved (for now) via https://github.com/esmero/format_strawberryfield/commit/51cd6e6d09a89fdaa803c613f2efce245ccf5b00

PR #204, opened 2022-06-03
ISSUE-203:OpenCV is the new Black

DiegoPino:
# What?

See #203 

Leaving this here and getting a coffee

image


aksm (approved):
Didn't review too closely since I don't want to delay merging, but testing functionality didn't reveal any major errors. @DiegoPino probably you've already noticed this, but the `FS error` goes away when I add a breakpoint and resume so looks like some kind of race condition? Will continue to debug to see if I can figure anything out.

format_strawberryfield.libraries.yml
```diff
@@ -13,11 +13,17 @@ iiif_openseadragon_strawberry:
...
   js:
     js/iiif-openseadragon_strawberry.js: {minified: false}
+    js/jquery_dollar.js: {preprocess: false, minified: false, weight: -10}
   dependencies:
     - core/jquery
+    - core/jquery.once
     - core/drupal
     - core/drupalSettings
+    - format_strawberryfield/iiif_openseadragon
     - format_strawberryfield/annotoriousopenseadragon
```
aksm, line 23:
```suggestion
    - format_strawberryfield/annotoriousopenseadragon
    - format_strawberryfield/opencv
```
Adding opencv as a dependency seems to get rid of the `FS error` in my testing.

PR #216, opened 2022-07-19
ISSUE-182: Make Metadata Display Entities Revisionable

### LINKED ISSUE #182: Make Metadata Display Entities Revisionable
# What?

Since metadata display entities drive almost every user facing and import facing transformation we should make them revision able so small changes/mistakes (logic ones, since we already validate them) can be reverted and also there exposed log of the evolution of a twig template in the form of these entities

# How?

The task at hand is well document 
- https://www.drupal.org/docs/drupal-apis/entity-api/converting-a-content-entity-type-to-be-revisionable-and-publishable

And involves:
- Updating the Annotations on the Entity
- Adding new Form routing for the Entity
- Extending the Entity to a different Class
- Making Update Hooks for already ingested Entities
An on our side/integration
- Making sure that we always use the latest published Entity whenever we load it (testing first, then checking where we fail)

Twig templates (our metadata display entities) are used on
- Format Strawberry Field
- Webform (not too much, probably a small check on the current webform elements is enough)
- AMI  
   
@aksm this should not take more than a week. Please open a pull as soon as you start working on this so we can have an ongoing discussion. Thanks!

aksm:
Still needs work, but begins to resolve #182

### DIFF
--- a/format_strawberryfield.install
+++ b/format_strawberryfield.install
@@ -270,6 +270,7 @@ function format_strawberryfield_update_8703(array &$sandbox) {
     ->setRevisionable(TRUE)
     ->setTranslatable(TRUE)
     ->setDefaultValue(TRUE)
+    ->setInitialValue(TRUE)
     ->setDisplayOptions('form', [
       'type' => 'boolean_checkbox',
       'settings' => [
--- a/format_strawberryfield.post_update.php
+++ b/format_strawberryfield.post_update.php
@@ -93,6 +93,7 @@ function format_strawberryfield_post_update_make_metadatadisplay_entity_revision
     ->setRevisionable(TRUE)
     ->setTranslatable(TRUE)
     ->setDefaultValue(TRUE)
+    ->setInitialValue(TRUE)
     ->setDisplayOptions('form', [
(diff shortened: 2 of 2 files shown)


format_strawberryfield.install
```diff
@@ -214,3 +214,128 @@ function format_strawberryfield_update_8702() {
...
+    ->setDisplayOptions('form', [
+      'type' => 'string_textarea',
+      'weight' => 25,
+      'settings' => [
+        'rows' => 4,
+      ],
+    ]);
+  $definition_update_manager->installFieldStorageDefinition('revision_log_message', $entity_type_id, $entity_type_id, $revision_log_message);
+  //$last_installed_schema_repository = \Drupal::service('entity.last_installed_schema.repository');
+  //$field_storage_definitions = $last_installed_schema_repository->getLastInstalledFieldStorageDefinitions('metadatadisplay_entity');
```
DiegoPino, line 318: @aksm why are these lines commented out?

aksm: Meant to delete those when I realized they needed to be in the post_update hook.

PR #225, opened 2022-09-22
ISSUE-224: Fix UUID Autocomplete validator when using Ajax (Views)

DiegoPino:
This is a stupid Views issue, not us. When loading a formatter for a Field in a Views that uses the entity UUID autocompleted we build (because you don't want to reference IDs, those change people), the Views opens an Ajax Windows that calls the validation of the Settings form before even displaying. At that moment the $value is NULL and you can't load an entity from storage with a NULL value. This small checks on nulls/elseif deals with it.  Also multiple entities should work now (but i have no use for that right now) Only issue is it will actually mark the field as "incomplete" but i don't see a way around this at all bc validation should not be called on Form render. Stupid views.


DiegoPino:
@aksm ping

aksm (approved):

Issue #128, opened 2021-02-18
XML HOCR/Alto support and alternative Mirador build using the dbmdz overlay work


DiegoPino:
# What is this?

Look kids: https://mirador-textoverlay.netlify.app
And source code  https://github.com/dbmdz/mirador-textoverlay

This is a custom Mirador 3 plugin build by Wizard and Techno connoisseur @jbaiter and @dbmdz team that allows selectable and renderable Overlays using IIIF Manifests and the 'seeAlso' key. There is a lot of work behind on making this computable and renderable in real time (a lot of browser debugging). Since we are already enjoying their work (and thankful) and code on HOCR highlighting directly from Solr inside our custom IABookreader extensions  I would love to integrate this as a secondary Mirador/Formatter for that use case.

Example from https://wellcomelibrary.org/iiif/b18035723/manifest

```JSON
"@id": "https://wellcomelibrary.org/service/alto/b18035723/0?image=0",
  "seeAlso": {
            "@id": "https://wellcomelibrary.org/service/alto/b18035723/0?image=0",
            "format": "text/xml",
            "profile": "http://www.loc.gov/standards/alto/v3/alto.xsd",
            "label": "METS-ALTO XML"
          },
```

# What is needed?

- We need to add a new dynamic Route/endpoint that can deliver our Solr Stored miniOCR as miniOCR.xml (obvious) but also transform on the fly (blow up) Alto and/or HOCR ? (and we may not need both). miniOCR is not yet supported but we are happy to adapt. This (fun) can either be a Twig template or an actual XSLT. Twig will process faster of course.

- Secondly we need to build our own Mirador 3 and add this and hopefully to use the energy a few more Plugins.I wish Plugins where dynamically but we will have to not use CDN for this. I mean eventually, once built, via the GitHub magic to CDN connection we will be able to serve that way.

- Adapt our IIIF manifest 3 Metadata Display Entities (Twig) to expose OCR when already present as documented using ` "seeAlso"` (or always even if not present yet)  

-Report back and show @dbmdz our love and implementation.

That should be all. This also means, we need to start injecting (looking at @alliomeria 👀 ) our other WebAnnotations, rectangles and polygons into the manifests.

Thanks!

giancarlobi:
@DiegoPino Great!! I think we need only one between ALTO or hOCR, probably the one nearest to miniOCR data.

DiegoPino:
[...]

PR #235, opened 2022-11-02
ISSUE-128: Basic pass custom Mirador 3.3.0 with Image tools

DiegoPino:
See #128

I could not make the awesome text-overlay the Johannes wrote play with image-tools that our community asked for. When both are enabled image tools are not drawn. I have no idea what is happening and that will imply extra work. Some funkyness with opening new manifests when pressing on the blue signs are also happening, might be some material-layour issue, a css that is missing, a selector or even the URL parser. JS -react- is a mystery for me.

Anyhow. This works. Will add a second JS with just text-select to cover that need and since we are loading this selectively I will enable that option too

Will also commit at the same time UV viewer here. That covers more needs. Will try to dig deeper in what is wrong with the super simple custom Webpack JS I built but I have the feeling it is an issue with one of the dependencies (2pm might have gotten something a BIT newer)


DiegoPino:
@patdunlavey seems like the problem (finally) is related to React version in the newer Mirador. The solution is to re-build all of this on a previous release

PR #238, opened 2022-11-07
ISSUE-236: Citation Formatter check for null

### LINKED ISSUE #236: Citation Formatter check for null
# What?
`$rendered_json_string = $metadatadisplayentity->renderNative($context);` in `StrawberryCitationFormatter.php` needs a check before in case the metadata display entity being loaded is NULL.

aksm:
Resolves #236 for branch 1.1.0.

The check for NULL was already there. Just needed to move the rest of the metadata display entity rendering into the try block.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,6 @@
         "erusev/parsedown": "^1.7",
         "drupal/webform": "^5.23 || ^6.0",
         "ext-json": "*",
-	"seboettg/citeproc-php": "master"
+	"seboettg/citeproc-php": "dev-master"
     }
 }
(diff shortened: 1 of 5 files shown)


DiegoPino:
Albert. Can you remove that typo/fix from this pull? I opened a new one for that (since you opened a new issue I feel that is cleaner, and it also only applies to 1.1.0, while this pull needs to go into 1.1.0 and 1.0.0)

Thanks!

Diego Pino Navarro
Director of Digital Strategy
Archipelago architect
Metropolitan New York Library Council
PO Box 2084
New York, NY 10108

DiegoPino (approved):
Good!

Issue #237, opened 2022-11-07
Spacing error in format_strawberryfield.schema.yml


aksm:
# What?
The below section throws an error because it should be aligned with `type: mapping` just above it:
https://github.com/esmero/format_strawberryfield/blob/e278f711d7ee015dab4e2da781eaf2db9b7b1983/config/schema/format_strawberryfield.schema.yml#L385-L403

patdunlavey:
Yep - just discovered this one too!

DiegoPino:
Thanks. Fixing.

aksm:
Resolved by 3af528457663395482f14c51cd428ce6be0053cd

PR #239, opened 2022-11-07
ISSUE-237: bad indentation for UV formatter schema

DiegoPino:
See #237

Please check if I got it right this time? @patdunlavey @aksm 

Thanks for reporting


aksm:
@DiegoPino Oops, I included the same fix in #238 just before you opened this. Should I back that out of there?

DiegoPino:
Yep. Different issues but also your bug fix needs to apply cleanly on 1.0.0 and 1.1.0 and 1.0.0 does not have the UV viewer at all

aksm (approved):

PR #243, opened 2022-11-08
ISSUE-241: Citation Formatter settings undefined variables

### LINKED ISSUE #241: Citation Formatter settings undefined variables
# What?
The `$citationstyles` variable and the `$localekey` variable aren't initiated as `NULL` so users get an `undefined` warning when initially selecting and saving the formatter for a field. My own stupid copypasta mistake, but a simple fix.

aksm:
Resolves #241 for the 1.1.0 branch.


DiegoPino:
matches #242 (which was for 1.0.0/ backwards fix)

Issue #244, opened 2022-11-08
Workaround Mirador to allow mutiple Audios on Chrome/Firefox


patdunlavey:
This is the exact same issue for audio files as @DiegoPino found for videos, and I found that the workaround devised for videos also works for audios. What the user sees is that when more than one audio file thumbnail is displayed in Mirador, you can click on the first one, but then clicking on a different audio thumbnail has no effect. I was able to make this work by duplicating Diego's solution in [mirador_strawberry.js](https://github.com/esmero/format_strawberryfield/blob/main/js/mirador_strawberry.js#L47) for audio elements, which I documented on the Mirador github issue [here](https://github.com/ProjectMirador/mirador/issues/3486#issuecomment-1307514086).

@DiegoPino would you like me to send you a PR based on that^ solution?

DiegoPino:
@patdunlavey please do so. The Mutation observer is nice for that and saw your reply. Please make a pull against 1.1.0? Thanks!

patdunlavey:
Will do!

DiegoPino:
Gracias!

DiegoPino:
Solved via https://github.com/esmero/format_strawberryfield/commit/370086ef09d775ff461ffddad50b5388c32ac216

Thanks @patdunlavey on a next iteration we try to optimize!

PR #245, opened 2022-11-08
ISSUE-244 --- Workaround Mirador to allow mutiple Audios on Chrome/Firefox

patdunlavey:
See https://github.com/esmero/format_strawberryfield/issues/244

I replicated Diego's [solution for video player links in Mirador](https://github.com/esmero/format_strawberryfield/commit/c823b3eb459bf1327619a0064dc9c3e9b951c6b2) to work for audio player links.

To test, create an ADO with multiple audio files and configure the ADO type to display using the Mirador viewer and IIIF Presentation 3.0 API.

### DIFF
--- a/js/mirador_strawberry.js
+++ b/js/mirador_strawberry.js
@@ -90,6 +90,14 @@
                                     }
                                 });
                             }
+                            let mirador_audios = document.querySelectorAll(".mirador-viewer audio source");
+                            if (mirador_audios.length) {
+                                mutations.forEach(function (mutation) {
+                                    if ((mutation.target.localName == "audio") && (mutation.addedNodes.length > 0) && (typeof(mutation.target.lastChild.src) != "undefined" )) {
+                                        mutation.target.src = mutation.target.lastChild.getAttribute('src');
+                                    }
+                                });
+                            }
                         });
                         observer.observe(mirador_window, {
                             childList: true,


patdunlavey:
@DiegoPino do we care that this is not efficient code-wise? Should I attempt to merge these two blocks of code into one that handles both video and auto elements?

js/mirador_strawberry.js
```diff
@@ -90,6 +90,14 @@
                                     }
                                 });
                             }
+                            let mirador_audios = document.querySelectorAll(".mirador-viewer audio source");
```
DiegoPino, line 93: Super cool! On a next iteration we can make a this condition

` if (mirador_videos.length ||  mirador_audios.length)`
and then a single `mutations.forEach(function (mutation)` (which is the most expensive for)
mutation.target.localName == "video" and mutation.target.localName == "audio" will work of course (or should?)

Just an idea for the future (or I will forget it)

DiegoPino:
@patdunlavey thanks! Give me a few minutes, will merge ASAP. Look at the tiny comment I added, I think we could save a few iterations in the future.

Also: Extra future idea (using mutations!)

See https://github.com/esmero/format_strawberryfield/blob/033a4a834fbc5c52980931a3bc6fa31abd92cf02/js/iiif-pannellum_strawberry.js#L229-L235

For pannellum, everytime I change a scene I trigger this event. The map knows how to "listen" to this.
Imagine we could do the same on every Canvas change ...!!

Would allow e.g interactive changes in other viewers based (metadata/transcripts/etc) on Canvas/resource loads.
Not saying it is easy, but might be a good future challenge. 

Thanks a lot for this!

DiegoPino (approved):
Good! Thanks for your contribution, super useful :)

DiegoPino:
> @DiegoPino do we care that this is not efficient code-wise? Should I attempt to merge these two blocks of code into one that handles both video and audio elements?

Jinx. I added a comment. Gosh are we relatives/did we go to school together!?

DiegoPino:
I will leave the "optimize here" or later up to you. Now would be great but I don't want to abuse your time

js/mirador_strawberry.js
patdunlavey: I was wondering if we could just expand the selector for both av types. Does this work (pardon my laughable jquery ignorance!!)?
```js
 let mirador_avs = document.querySelectorAll(".mirador-viewer video source, .mirador-viewer audio source");
```
then...
```js
if ($.inArray(mutation.target.localName, ["audio","video"]) && (mutation.addedNodes.length > 0) && (typeof(mutation.target.lastChild.src) != "undefined" )) {
```

patdunlavey:
If you're ok with merging, I am too!

js/mirador_strawberry.js
DiegoPino: I guess so? What if I merge, and we try this first? =)

patdunlavey:
Pleasure doing business with you sir!

Issue #249, opened 2022-11-16
Ado type condition plugin breaks block visibility form


patdunlavey:
When editing a block, the `required = TRUE` attribute for the  'ado_type' form element causes form submit to fail, with `An invalid form control with name='visibility[ado_type_condition][ado_types]' is not focusable.` message in the console. It also causes a context condition form to fail to submit if the "ADO Type" is not entered and the tab is not visible.

I think the solution is simply to remove [line 65 of Plugin/Condition/AdoType.php](https://github.com/esmero/format_strawberryfield/blob/1.1.0/src/Plugin/Condition/AdoType.php#L65). There is no purpose that I can think of in making this field required. If you fill it in, great, if not, then the evaluate condition will normally return true (unless you check "negated"). So I don't think there's any harm in just removing the "#required" attribute.

Do you agree @DiegoPino ? Shall I make a PR?

patdunlavey:
PR Created: https://github.com/esmero/format_strawberryfield/pull/250

DiegoPino:
Fixed via https://github.com/esmero/format_strawberryfield/commit/fa73dba305ee2bfb1c8f7cfc7f9a709368cb702a
Thanks @patdunlavey!

patdunlavey:
Thanks @DiegoPino !!

PR #250, opened 2022-11-16
ISSUE-249 --- removed '#required' from the 'ado_types' field in the ADOType condition configuration form

patdunlavey:
See https://github.com/esmero/format_strawberryfield/issues/249

Requiring the 'ado_types' field to be required causes drupal's block configuration forms to become unsubmittable, and context conditions to sometimes become unsubmittable if the form tab is hidden when the user tries to submit the form. This PR solves the problem simply by removing the "required" attribute, since `format_strawberryfield\Plugin\Condition\AdoType::evaluate` handles the case of this field being empty just fine.

### DIFF
--- a/src/Plugin/Condition/AdoType.php
+++ b/src/Plugin/Condition/AdoType.php
@@ -113,8 +113,15 @@ class AdoType extends ConditionPluginBase implements ContainerFactoryPluginInter
         }
       }
     }
+    else {
+      // This not an entity type that bears a strawberryfield. Therefore, this condition can not apply.
+      // However, it could still be negated and cause in a "TRUE" evaluation to flip and result in the "FALSE"
+      // being returned. Deal with that by removing the negation.
+      unset($this->configuration['negate']);
+      return TRUE;
+    }
     if(!empty($ado_types)) {
-      // Return true if the any of the entity's types are in the condition's ado_types.
+      // Return true if any of the entity's types are in the condition's ado_types.
       return (count(array_intersect($this->configuration['ado_types'], $ado_types)) > 0);
     }
     // Default, return not matched.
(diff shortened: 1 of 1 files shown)


DiegoPino:
You sure https://github.com/esmero/format_strawberryfield/blob/e7247513a4e53e1c301001e64045833102b6b19a/src/Plugin/Condition/AdoType.php#L118 will not fail if empty? Empty means an empty array?

Could you please double check the evaluate logic

patdunlavey:
Thanks for looking at this @DiegoPino . I'm confident that we can't have a variable type mismatch for $this->configuration['ado_types']. It's always an array.

I did notice that it does not handle the case where the condition is applied to a non-strawberryfield-bearing entity correctly. It should always return true in that case. Additionally, it should remove the 'negate' condition in that case, so that a true does not end up being reversed to a false.

DiegoPino:
@patdunlavey do you want to tackle those issues in this pull? Let me know if not I give it a try

patdunlavey:
Sorry, I failed to mention that I updated the pr to include my attempt at those fixes! Could you re-review?

DiegoPino:
All good! Will review now. Thanks!

src/Plugin/Condition/AdoType.php
```diff
@@ -114,8 +113,15 @@ public function evaluate() {
         }
       }
     }
+    else {
+      // This not an entity type that bears a strawberryfield. Therefore, this condition can not apply.
+      // However, it could still be negated and cause in a "TRUE" evaluation to flip and result in the "FALSE"
+      // being returned. Deal with that by removing the negation.
+      unset($this->configuration['negate']);
```
DiegoPino, line 120: Oh! Learned a thing here. So removing it means solves this because of
```
public function isNegated() {
    return !empty($this->configuration['negate']);
  }
```

DiegoPino (approved):
Ok, reads well to me!

patdunlavey:
Thanks @DiegoPino !!

PR #257, opened 2022-12-10
Small improvement to `IIIF_V3_JMESPATH`

digitaldogsbody:
A small optimisation to the V3 path that avoids the need to store `@id`/`id` and `@type`/`type` in temporary hashes, and just assigns them straight away.

Playground: https://play.jmespath.org/?u=5ad1f201-12d6-46aa-96bf-9174c6541a7a


DiegoPino (approved):
Thanks! Looks very good to me

PR #266, opened 2023-02-02
ISSUE-265: Allow multiple CSS classes for clipboard copy functionality

### LINKED ISSUE #265: Allow multiple CSS classes for clipboard copy functionality
# What?
The javascript for clipboard functionality currently only accepts one CSS class for the button that is created. It would be better to allow multiple classes so Bootstrap classes can be added for easier theming.

aksm:
Resolves #265. The data attribute used for the copy button's CSS classes can contain a space-separated list. Once this is merged 2 small updates can follow:

1. Open PR for documentation around this copy/paste functionality should be updated.
2. Javascript for the IA Bookreader share functionality can be slightly refactored to be cleaner.

### DIFF
--- a/js/clipboard_copy_strawberry.js
+++ b/js/clipboard_copy_strawberry.js
@@ -26,20 +26,27 @@
           if(copyButtonContainers !== undefined) {
             for(var i = 0; i < copyButtonContainers.length; i++) {
               var copyButtonContainer = copyButtonContainers[i];
-              var copyContentContainer = copyContentContainers[i];
-              if(!copyContentContainer.hasAttribute('id')) {
-                copyContentContainer.id = 'clipboard-copy-content-' + i;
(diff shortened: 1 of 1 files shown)


DiegoPino:
@aksm @alliomeria @karomabiles since you are here now, could we please talk about the HTML being generated? I feel it could really need some improvement to allow more styling options. Team call?

aksm:
@DiegoPino this works with any user-defined content source and custom button, but it will require clear documentation, which I'll follow up with shortly.

DiegoPino (approved):
@aksm approving this. Could you share somewhere a small Twig snipped? I think that would serve as proper documentation. Once that is done I will merge ASAP. thanks!

aksm:
Hi @DiegoPino! I've added to the pending documentation here: <https://github.com/esmero/archipelago-documentation/pull/152/commits/531854032267078fb49a0572de3c2d8e9060985c>

PR #268, opened 2023-02-09
ISSUE-223: Provide media/files at UUIDs additionally to node IDs

### LINKED ISSUE #223: Provide media/files at UUIDs additionally to node IDs
# What
Related to esmero/strawberryfield#220 and esmero/strawberryfield#230.

When downloading media/files for ADOs they are currently accessed at unstable node IDs. As we do for the ADOs themselves, they should be provided at stable UUIDs.

The suggested path will be as follows:
`/do/{uuid}/file/{uuid}/download/{format}`

@DiegoPino Not sure if I'm describing this correctly so please correct me if I'm wrong or need to clarify.

aksm:
Resolves #223.

### DIFF
--- a/format_strawberryfield.routing.yml
+++ b/format_strawberryfield.routing.yml
@@ -66,7 +66,7 @@ format_strawberryfield.iiifbinary:
     _entity_access: 'node.view'
 
 # Direct File access for SBF managed files via UUID.
-format_strawberryfield.iiifbinary_uuid:
+format_strawberryfield.binary:
   path: '/do/{node}/file/{uuid}/download/{format}'
   methods: [GET,HEAD]
   defaults:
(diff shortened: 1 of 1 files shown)


format_strawberryfield.routing.yml
```diff
@@ -65,6 +65,23 @@ format_strawberryfield.iiifbinary:
     format: .+
     _entity_access: 'node.view'
 
+# Direct File access for SBF managed files via UUID.
+format_strawberryfield.iiifbinary_uuid:
```
DiegoPino, line 69: Hi, even if for legacy reasons the controller class is named IiifBinaryController, it would be better to name the route simply format_strawberryfield.binary maybe?

DiegoPino:
@aksm thanks. I will test tonight and merge!

DiegoPino (approved):
Good!

PR #277, opened 2023-03-03
New submodule: format_strawberryfield_rest_oai_pmh

patdunlavey:
This PR provides a new submodule for format_strawberryfield, entitled format_strawberryfield_rest_oai_pmh, which basically lets you use metadatadisplay templates to perform the transformation from ADO metadata to a form that can be exposed for harvesting via the endpoint configured in the rest_oai_pmh module.
The usage instructions (copied here from the readme) are:
- Enable this module and it's dependencies (format_strawberryfield, rest_oai_pmh).
- Create a view with an entity reference display that selects the ADOs that you wish to be harvested through OAI. Consult rest_oai_pmh module instructions for details. Note that there is a bug in rest_oai_pmh that prevents it being used with search_api_solr views. See below for a workaround until that bug is fixed.
- Create metadatadisplay templates for MODS and/or Dublin Core that transform the native strawberryfield json into valid xml metadata that you wish to be harvested.
- Go to "Administration > Configuration > Web services > Rest > REST OAI-PMH Settings > Templates" and select the metadatadisplay templates created in the previous step for MODS and/or DC.
- At "Administration > Configuration > Web services > Rest > REST OAI-PMH Settings", under "What to expose to OAI-PMH", select the entity reference view you created earlier. In the "Metadata Mappings", select "MODS Metadatadisplay Template" and/or "Dublin Core Metadatadisplay Template".

As noted, there is a bug in rest_oai_pmh that prevents it from working with search_api_solr views. I have an [MR](https://git.drupalcode.org/project/rest_oai_pmh/-/merge_requests/9) for that, but the module maintainer seems to be being slow to respond. (Maybe you could review that @DiegoPino and offer any of your own suggestions?)


DiegoPino:
@patdunlavey this is awesome. Will check that open MR on the base module, I think the issue there is that he is using non generic methods to access the entity/type. The search API itself has an abstraction and thus might be possible to cut your PR to a bit of less code. I wonder if we could ship this module here with your pull as patch?

Give me the weekend, this is one I will put some good hours to make sure its merged by Monday. Thanks so much!!

patdunlavey:
Joecorall merged my MR over the weekend. He [wrote](https://www.drupal.org/project/rest_oai_pmh/issues/3344748#comment-14951817): "Thanks for the MR. I merged it into the dev version. I am going to test this for the known node case. The code was designed to possibly support entity types other that nodes, but to my knowledge that use case never presented itself until what looks like now. So the other entity type support with solr was more theoretical. I am going to make sure this doesn't cause any regressions then cut a new release."

I replied that "a colleague" had suggested there might be a more efficient way to determine the entity type and entity id for each record. So if you have time to give me some ideas about that, I would be grateful!

patdunlavey:
I updated the README to remove the instructions for dealing with the fork of rest_oai_pmh.

DiegoPino:
@patdunlavey Ok, so caching is not bubbling but then this is MODS and DC only and dealing with that (needed for Views inclusions and dynamic stuff) is not a larger concern. I can see how we can work on this deeper in the future by allowing, instead of fixed MODS/DC a deriver plugin handler that could also allow the wrapper to be provided by either config OR a twig template too. Not sure about the rest_oai_pmh module and its caching strategy though (but that is a design issue out of our control here).

So.. approving and welcoming a new module to the Archipelago community. Thanks!

DiegoPino (approved):
Ready for a merge. Let's see how this performs in reality!

patdunlavey:
Thanks Diego!!!

Issue #276, opened 2023-03-02
Add/Remove for Advanced Search doesn't work when the operator is not exposed


patdunlavey:
See https://github.com/esmero/format_strawberryfield/issues/260#issuecomment-1452142415.

When the exposed advanced search form is built/rebuilt [here](https://github.com/esmero/format_strawberryfield/blob/2e5cc2a6f882cd2490caccc2cbd5b88b58f70c0f/modules/format_strawberryfield_views/src/Plugin/views/filter/AdvancedSearchApiFulltext.php#L667), it only adds/removes filter fields if the form elements are contained inside a wrapper div. 

However, the wrapper element is only created "When the operator and value forms are both in play": https://git.drupalcode.org/project/drupal/-/blob/10.1.x/core/modules/views/src/Plugin/views/filter/FilterPluginBase.php#L961.

This task is to make add/remove work, whether or not the exposed filter operator is exposed.

DiegoPino:
Fixed via https://github.com/esmero/format_strawberryfield/commit/b3571c01558493abde3af2b92e8b6cfc9406be9f by @patdunlavey . Thanks

PR #278, opened 2023-03-03
Fix Issue 276: advanced search add/remove without exposed operator

patdunlavey:
See https://github.com/esmero/format_strawberryfield/issues/276

The wrapper fieldset is needed for the add/remove mechanism to work. This PR ensures that it will be there, whether or not it started out that way.

### DIFF
--- a/modules/format_strawberryfield_views/src/Plugin/views/filter/AdvancedSearchApiFulltext.php
+++ b/modules/format_strawberryfield_views/src/Plugin/views/filter/AdvancedSearchApiFulltext.php
@@ -578,15 +578,18 @@ class AdvancedSearchApiFulltext extends SearchApiFulltext {
         '#default_value' => $form_state->getValue($searched_fields_identifier),
       ];
 
-      if (isset($form[$this->options['id'] . '_wrapper'])) {
-        $form[$this->options['id'] . '_wrapper'][$searched_fields_identifier]
-          = $newelements[$searched_fields_identifier];
+      if (empty($form[$this->options['id'] . '_wrapper'])) {
+        $form[$this->options['id'] . '_wrapper'] =
+          [
+            "#type" => "fieldset",
+            "#title" => t('Advanced Fulltext search'),
+            $this->options['id'] => $form[$this->options['id']],
+           ];
+        unset($form[$this->options['id']]);
+      }
+      $form[$this->options['id'] . '_wrapper'][$searched_fields_identifier]
+        = $newelements[$searched_fields_identifier];
 
-      }
-      else {
-        $form[$searched_fields_identifier]
-          = $newelements[$searched_fields_identifier];
-      }
     }
 
     $advanced_search_operator_id = $this->options['id'] . '_group_operator';
@@ -611,12 +614,7 @@ class AdvancedSearchApiFulltext extends SearchApiFulltext {
         '#access' => FALSE,
         '#default_value' => $form_state->getValue($advanced_search_operator_id,'or'),
       ];
-      if (isset($form[$this->options['id'].'_wrapper'])) {
-        $form[$this->options['id'] . '_wrapper'][$advanced_search_operator_id] = $newelements[$advanced_search_operator_id];
-      }
-      else {
-        $form[$advanced_search_operator_id] = $newelements[$advanced_search_operator_id];
-      }
+      $form[$this->options['id'] . '_wrapper'][$advanced_search_operator_id] = $newelements[$advanced_search_operator_id];
     }
 
     // Yes over complicated but so far the only way i found to keep the state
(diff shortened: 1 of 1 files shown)


DiegoPino:
@patdunlavey interesting approach. This feels good. Will test with your view and ours and will get back to you tomorrow AM. Thanks so much for putting this work in.

modules/format_strawberryfield_views/src/Plugin/views/filter/AdvancedSearchApiFulltext.php
```diff
@@ -578,15 +578,18 @@ public function buildExposedForm(&$form, FormStateInterface $form_state) {
         '#default_value' => $form_state->getValue($searched_fields_identifier),
       ];
 
-      if (isset($form[$this->options['id'] . '_wrapper'])) {
-        $form[$this->options['id'] . '_wrapper'][$searched_fields_identifier]
-          = $newelements[$searched_fields_identifier];
+      if (empty($form[$this->options['id'] . '_wrapper'])) {
+        $form[$this->options['id'] . '_wrapper'] =
+          [
```
DiegoPino, line 583: @patdunlavey small thing, can you move the `[` to the previous line? DCS

modules/format_strawberryfield_views/src/Plugin/views/filter/AdvancedSearchApiFulltext.php
```diff
@@ -578,15 +578,18 @@ public function buildExposedForm(&$form, FormStateInterface $form_state) {
...
+            "#type" => "fieldset",
```
DiegoPino, line 584: @patdunlavey this here should better come from the original Method, in case they decide to change how the wrapper is build (at the extended Filter class)

protected function buildValueWrapper(&$form, $wrapper_identifier) {
    // If both the field and the operator are exposed, this will end up being
    // called twice. We don't want to wipe out what's already there, so if it
    // exists already, do nothing.
    if (!isset($form[$wrapper_identifier])) {
      $form[$wrapper_identifier] = [
        '#type' => 'fieldset',
      ];

      $exposed_info = $this->exposedInfo();
      if (!empty($exposed_info['label'])) {
        $form[$wrapper_identifier]['#title'] = $exposed_info['label'];
      }
      if (!empty($exposed_info['description'])) {
        $form[$wrapper_identifier]['#description'] = $exposed_info['description'];
      }
    }
  }


So basically
$wrapper = $form[$this->options['id'] . '_wrapper'];
$this->buildValueWrapper($form, $wrapper);

That also makes the code slimmer. What do you think?

patdunlavey: Does this fix it?

modules/format_strawberryfield_views/src/Plugin/views/filter/AdvancedSearchApiFulltext.php
patdunlavey: This line is removed by using the buildValueWrapper method (see next)

modules/format_strawberryfield_views/src/Plugin/views/filter/AdvancedSearchApiFulltext.php
```diff
@@ -578,18 +578,13 @@ public function buildExposedForm(&$form, FormStateInterface $form_state) {
...
       ];
 
-      if (empty($form[$this->options['id'] . '_wrapper'])) {
-        $form[$this->options['id'] . '_wrapper'] =
-          [
-            "#type" => "fieldset",
-            "#title" => t('Advanced Fulltext search'),
-            $this->options['id'] => $form[$this->options['id']],
-           ];
+      if (empty($form[$this->options['id'] . '_wrapper']) && !empty($form[$this->options['id']])) {
```
DiegoPino, line 581: Almost. I mean it does work, but if you check the method you don't need too check if the wrapper is already there or not, the  $this->buildValueWrapper() does it already for you.

patdunlavey: I put in the check because in the next line we make this assignment:
```
$form[$this->options['id'] . '_wrapper'][$this->options['id']] = $form[$this->options['id']];
```
If the form has already had the wrapper provided (e.g. if you elect to have exposed the operator), then `$form[$this->options['id']]` will be empty and would cause an error.

patdunlavey: Ok, I realize the mistake in my argument. Can we always assume the absence of the wrapper element guarantees that `$form[$this->options['id']]` exists? `buildValueWrapper` does not, by itself, guarantee it.

patdunlavey: Ok - removed.

DiegoPino (approved):
Hi, thanks a lot. Not exactly what I meant but it does work and for the sake of moving forward I I approve and will merging. I will come back to this after that. Thanks so much for your work on this

patdunlavey:
Thanks very much. Sorry for the unresolved confusion! Sometimes, you just gotta spell it out!

Issue #284, opened 2023-03-22
Advanced Full text Search Views Filter plugin not using the right/per field operator


DiegoPino:
# What?

This was reported yesterday after checking it is a bug in my code. Views passes the Operator for either the complete Filter/Views or in our case the Specific Advanced Full Text Filter as lowercase and Solr Search API mode parser which uses a complex array of arrays needs Uppercase and if not will default to "AND". Because I'm preparing the Fields/Values combo through the "chosen/configured" parser to build a very hackie (needed bc Direct Queries are deleted by the Solr Search API so I have to mimic also that it is DISMAX and then remove the dismax flag) Direct Query I missed this specific part.

## Solution

Simple, uppercase the hell of this. Single line + a comment

@alliomeria @patdunlavey (because you will need this)

DiegoPino:
Fixed via https://github.com/esmero/format_strawberryfield/commit/b39038f65be93747a1a9878af89483ee63f4b2eb

PR #285, opened 2023-03-22
ISSUE-284: Uppercase the Conjunction Operator to please the Solr Gods (rey sol)

DiegoPino:
See #284 

Fixes the bug by ensuring that Operator a.k.a conjunction is passed in Solr Search API flavor (uppercased). Also makes sure that we always limit the operator to or/and/not and then upper case. This also forces me to do the #negation against the uppercase version


aksm (approved):

DiegoPino:
Thanks @aksm !

Issue #283, opened 2023-03-20
Facet Summary Processor misbehavior on Date Range + Search Query


DiegoPino:
# What?

For some strange reason (reason I'm opening this ISSUE) when a Facet Range Widget + a normal Search Query is done and there are no results the Facet Summary Links for both the date Facet entry and the Search item are incorrect and do not allow to "disable" just one of them.

What is strange is if we add an extra facet selection to this (even without results) the Links are OK.

This is happening somewhere here:
https://github.com/esmero/format_strawberryfield/blob/51a827bd82bbd8e9f591e757acca8cace41e6828/modules/format_strawberryfield_facets/src/Plugin/facets_summary/processor/LastActiveFacetsProcessor.php#L32-L269

## How to test (make it fail)?

Enable this Facet summary processor with all its options and disable the existing "Show a message when no results" built in one. Make a Faceted Search using a Date Range Facet. Then search for something that you know does not exist. When no results are returned this processor will try to generate Facet Summaries using an alternative method but the links for each Facet Summary will be wrong. 

## How to see the odd part working?
Enable this Facet summary processor with all its options and disable the existing "Show a message when no results" built in one. Make a Faceted Search using a Date Range Facet. The select another Facet (e.g Digital Object Type). Then search for something that you know does not exist. When no results are returned this processor will try to generate Facet Summaries using an alternative method and the links will all be right allowing you to deselect one of the choices without affecting the others.

PR #286, opened 2023-03-23
ISSUE-283:Facet Summary Processor misbehavior on Date Range when it is the only Facet selected

DiegoPino:
See #283 

This one was tricky. Gist: To generate a "unselect the current facet" you have to build all active facets except the one you want to unselect. When there are no ones (means the current, in this case our custom range) Facet Module suggests (copied the code) to use a HTTP REQUEST based parser method that gets confused when a single View (Facet Source) has multiple Displays (specially if those display are not 1:1).

The solution was not terrible.

Now that this is working I also made "quoting" the Full Text Search Terms Facet Summary (part of this pull)

### DIFF
--- a/modules/format_strawberryfield_facets/config/schema/format_strawberryfield_facets.schema.yml
+++ b/modules/format_strawberryfield_facets/config/schema/format_strawberryfield_facets.schema.yml
@@ -84,7 +84,7 @@ plugin.plugin_configuration.facets_summary_processor.sbf_last_active_facets:
       label: 'Enable Full text search terms as Facet Summary Entry'
     quote_query:
       type: boolean
-      label: 'If Fulll Text search terms will be surrounded by Double Quotes or not'
+      label: 'If Full Text search terms will be surrounded by Double Quotes or not'
     text:
       type: mapping
       mapping:
--- a/modules/format_strawberryfield_facets/config/schema/format_strawberryfield_facets.schema.yml
+++ b/modules/format_strawberryfield_facets/config/schema/format_strawberryfield_facets.schema.yml
@@ -15,6 +15,17 @@ facet.widget.config.sbf_date_range:
       type: integer
       label: 'Hide Reset when no selection'
 
+plugin.plugin_configuration.facets_processor.sbf_date_range:
+  type: mapping
+  label: 'Format Strawberryfield Date Range Picker'
+  mapping:
+    enabled:
+      type: integer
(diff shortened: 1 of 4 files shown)


DiegoPino:
@aksm if you have a few minutes, could you check if you see some non-sense here? Code is tested but maybe I missed a comment, indentation, etc or a piece of the logic is bonkers! Thanks

modules/format_strawberryfield_facets/config/schema/format_strawberryfield_facets.schema.yml
```diff
@@ -64,13 +75,16 @@ plugin.plugin_configuration.facets_summary_processor.sbf_last_active_facets:
...
       type: integer
-      label: Enable Empty Message
+      label: 'Enable Empty Message'
     enable_query:
       type: integer
-      label: Enable Full text search terms as Facet Summary Entry
+      label: 'Enable Full text search terms as Facet Summary Entry'
+    quote_query:
+      type: boolean
+      label: 'If Fulll Text search terms will be surrounded by Double Quotes or not'
```
aksm, line 87:
```suggestion
      label: 'If Full Text search terms will be surrounded by Double Quotes or not'
```

modules/format_strawberryfield_facets/src/Plugin/facets/widget/DateRangeWidget.php
```diff
@@ -227,7 +227,7 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
     ];
     $form += parent::buildConfigurationForm($form, $form_state, $facet);
 
-    $message = $this->t('To achieve the standard behavior of a Format Strawberryfield Date Range Picker, you need to enable the facet setting below <em>"Format Strawberryfield Date Range Picker"</em>.');
+    $message = $this->t('The Format Strawberryfield Date Range Picker Widget requires you to check the facet setting below <em>"Format Strawberryfield Date Range Picker" and inside it check "enabled" it.</em>.');
```
aksm, line 230:
```suggestion
    $message = $this->t('The Format Strawberryfield Date Range Picker Widget requires you to check the facet setting below <em>"Format Strawberryfield Date Range Picker" and check "enabled" inside it.</em>.');
```

aksm (changes requested):

DiegoPino:
@aksm thanks! I committed your change requests/suggestions

PR #288, opened 2023-03-29
ISSUE-287: Adds a new JOIN filter to allow Flavors to matching ADO unified search

DiegoPino:
This is WIP but works. Why WIP? The settings are not all the ones i need And also i want to expose (checkboxes) if this is enabled or not. It connects with/ Requires https://github.com/esmero/strawberryfield/pull/258 to enabled Highlight after the match was made

I could also work on improving this by 
- Exposing/Setting which Flavors to filter against (ONLY OCR?). For now we go for all which makes sense given that for one or another reason we are already indexing those right? (VTTs, XMLs, JSONs, HTML from webpages, you name it)
- Expose an ON/OFF switch/exposed filter on a view
- Making the Joined Solr fields able to `dvWithScore` or `topLevelDV` (last one allows millions to one) using `score=none` option to make joins even faster. This requires enabling other fields to be used to join ADOs and Flavors together (e.g the UUID instead of the ID which is `Int` using  `string`

### DIFF
--- a/modules/format_strawberryfield_views/format_strawberryfield_views.module
+++ b/modules/format_strawberryfield_views/format_strawberryfield_views.module
@@ -174,6 +174,18 @@ function format_strawberryfield_views_views_data_alter(array &$data) {
       if ($advanced_fulltext_field != 'sbf_advanced_search_api_fulltext') {
         $table[$advanced_fulltext_field]['real field'] = 'sbf_advanced_search_api_fulltext';
       }
+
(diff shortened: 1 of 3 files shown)


DiegoPino:
@aksm I might merge this later tonight but because this is not ready we can still make amends/more fixes tomorrow and the rest of the week under the same ISSUE. The exposed options + the idea of using strings implies a change in how the JOIN query is build but also a Schema change (extra fields) maybe?

DiegoPino:
@aksm @alliomeria a second pull will follow in the afternoon that drives what the Advanced Search needs to allow this

PR #296, opened 2023-04-17
ISSUE-289: Misreported unused JSON keys for multilevel keys in Metadata Display preview

### LINKED ISSUE #289: Misreported unused JSON keys for multilevel keys in Metadata Display preview
# What?

Following #275, a key such as `data.date_created_edtf.date_free` implies that `data.date_created_edtf` is used, but that is currently reported incorrectly.

Thanks to @alliomeria and @DiegoPino for noticing/reporting.

aksm:
Starts to resolve #289.

### DIFF
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -15,6 +15,10 @@ use InvalidArgumentException;
 use Drupal\format_strawberryfield\MetadataDisplayInterface;
 use Drupal\user\UserInterface;
 use Twig\Source;
+use Twig\Node\Expression\NameExpression;
+use Twig\Node\Expression\Unary\NotUnary;
+use Twig\Node\Expression\ConstantExpression;
+use Twig\Node\Expression\GetAttrExpression;
 
 /**
  * Defines the Metadata Display Content entity.
@@ -448,7 +452,7 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
       $lineno = [$node->getTemplateLine()];
       $variable_key = '';
       $parent_path = '';
-      if ($node instanceof \Twig\Node\Expression\NameExpression && !$nodes instanceof \Twig\Node\Expression\Unary\NotUnary) {
+      if ($node instanceof NameExpression && !$nodes instanceof NotUnary) {
         $variable_key = $node->getAttribute('name');
         if($variable_key == '_key') {
           $seq = $nodes->getNode('seq');
@@ -458,16 +462,16 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
           $parent_path = empty($seq_value) ? $seq_name : $seq_name . '.' . $seq_value;
         }
       }
-      elseif ($node instanceof \Twig\Node\Expression\ConstantExpression && $nodes instanceof \Twig\Node\Expression\GetAttrExpression) {
+      elseif ($node instanceof ConstantExpression && $nodes instanceof GetAttrExpression) {
         $variable_key = $node->getAttribute('value');
       }
-      elseif ($node instanceof \Twig\Node\Expression\GetAttrExpression) {
+      elseif ($node instanceof GetAttrExpression) {
         $variable_names = $this->getTwigVariableNames($node, array_replace_recursive($all_variables,$variables));
         $variable_key = implode('.', array_map(function($name) {
           return $name['path'];
         }, $variable_names));
       }
-      elseif ($node instanceof \Twig\Node\Node) {
+      elseif ($node instanceof Node) {
         $add_variables = $this->getTwigVariableNames($node, array_replace_recursive($all_variables,$variables));
         $variables = array_replace_recursive($variables, $add_variables);
       }
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -415,9 +415,9 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
   /**
    * {@inheritdoc}
    */
-  public function getTwigVariablesUsed() {
+  public function getTwigVariablesUsed(array $twigtemplate) {
     if ($this->usedTwigVars == NULL) {
-      $twigtemplate = $this->get('twig')->getValue();
+      $twigtemplate = empty($twigtemplate) ? $this->get('twig')->getValue() : $twigtemplate;
       $twigtemplate = !empty($twigtemplate) ? $twigtemplate[0]['value'] : "{{ 'empty' }}";
       // Create a \Twig Source first.
       $source = new Source($twigtemplate, $this->label(), '');
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -483,10 +483,10 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
         foreach ($variable_names as $name) {
           $name_check = $prev_name ? $prev_name['path'] . '.' . $name['path']:'';
           if (($name['path'] == $set_var) && !($set_source == $name_check)) {
-            array_push($variable_names_flat, $set_source);
+            $variable_names_flat[] = $set_source;
           }
           else {
-            array_push($variable_names_flat, $name['path']);
+            $variable_names_flat[] = $name['path'];
           }
           $prev_name = $name;
         }
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -443,6 +443,9 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
    * @param \Twig\Node\Node $nodes
    *   A Twig Module Nodes object.
    *
+   * @param array $all_variables
+   *   An array to track the variables during recursion to return the accumulated line numbers.
+   *
    * @return array
    *   A list of used $variables by this template.
    */
@@ -466,13 +469,13 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
         $variable_key = $node->getAttribute('value');
       }
       elseif ($node instanceof GetAttrExpression) {
-        $variable_names = $this->getTwigVariableNames($node, array_replace_recursive($all_variables,$variables));
+        $variable_names = $this->getTwigVariableNames($node, array_replace_recursive($all_variables, $variables));
         $variable_key = implode('.', array_map(function($name) {
           return $name['path'];
         }, $variable_names));
       }
       elseif ($node instanceof Node) {
-        $add_variables = $this->getTwigVariableNames($node, array_replace_recursive($all_variables,$variables));
+        $add_variables = $this->getTwigVariableNames($node, array_replace_recursive($all_variables, $variables));
         $variables = array_replace_recursive($variables, $add_variables);
       }
       if (!empty($variable_key) && is_string($variable_key)) {
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -446,10 +446,16 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
    * @param array $all_variables
    *   An array to track the variables during recursion to return the accumulated line numbers.
    *
+   * @param array $set_var
+   *   A string to track variables that reference JSON keys.
+   *
+   * @param array $set_source
+   *   An string to track the referenced JSON key for the above.
+   *
    * @return array
    *   A list of used $variables by this template.
    */
-  private function getTwigVariableNames(ModuleNode|Node|BodyNode $nodes, array $all_variables): array {
+  private function getTwigVariableNames(ModuleNode|Node|BodyNode $nodes, array $all_variables, string $set_var = '', string $set_source = ''): array {
     $variables = [];
     foreach ($nodes as $node) {
       $lineno = [$node->getTemplateLine()];
@@ -463,19 +469,26 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
           $seq_value = $seq->hasNode('attribute') ? $seq->getNode('attribute')->getAttribute('value') : '';
           $variable_key = $nodes->getNode('value_target')->getAttribute('name');
           $parent_path = empty($seq_value) ? $seq_name : $seq_name . '.' . $seq_value;
+          $set_var = $variable_key;
+          $set_source = $parent_path;
         }
       }
       elseif ($node instanceof ConstantExpression && $nodes instanceof GetAttrExpression) {
         $variable_key = $node->getAttribute('value');
       }
       elseif ($node instanceof GetAttrExpression) {
-        $variable_names = $this->getTwigVariableNames($node, array_replace_recursive($all_variables, $variables));
-        $variable_key = implode('.', array_map(function($name) {
-          return $name['path'];
+        $variable_names = $this->getTwigVariableNames($node, array_replace_recursive($all_variables, $variables), $set_var, $set_source);
+        $variable_key = implode('.', array_map(function ($name) use ($set_var, $set_source) {
+          if ($name['path'] == $set_var) {
+            return $set_source;
+          }
+          else {
+            return $name['path'];
+          }
         }, $variable_names));
       }
       elseif ($node instanceof Node) {
-        $add_variables = $this->getTwigVariableNames($node, array_replace_recursive($all_variables, $variables));
+        $add_variables = $this->getTwigVariableNames($node, array_replace_recursive($all_variables, $variables), $set_var, $set_source);
         $variables = array_replace_recursive($variables, $add_variables);
       }
       if (!empty($variable_key) && is_string($variable_key)) {
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -499,8 +499,7 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
       if (!empty($variable_key)) {
         $variables[$variable_key]['path'] = $variable_key;
         if(isset($all_variables[$variable_key]['line'])) {
-          $variables[$variable_key]['line']     = array_unique(array_merge($all_variables[$variable_key]['line'], $lineno));
-          $all_variables[$variable_key]['line'] = array_unique(array_merge($all_variables[$variable_key]['line'], $lineno));
+          $variables[$variable_key]['line'] = $all_variables[$variable_key]['line'] = array_unique(array_merge($all_variables[$variable_key]['line'], $lineno));
         }
         if(!isset($variables[$variable_key]['line'])) {
           $variables[$variable_key]['line'] = $lineno;
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -496,7 +496,7 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
         $add_variables = $this->getTwigVariableNames($node, array_replace_recursive($all_variables, $variables), $set_var, $set_source);
         $variables = array_replace_recursive($variables, $add_variables);
       }
-      if (!empty($variable_key) && is_string($variable_key)) {
+      if (!empty($variable_key)) {
         $variables[$variable_key]['path'] = $variable_key;
         if(isset($all_variables[$variable_key]['line'])) {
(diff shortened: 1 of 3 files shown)


DiegoPino:
@aksm, thanks. Will do a thorough code review tomorrow AM. I will also check of  ::arrayToFlatCommonkeys here (and others) can not be reused https://github.com/esmero/strawberryfield/blob/69c1c4e7e44572a548ba6374731f4db66976f295/src/Tools/StrawberryfieldJsonHelper.php, seems like there is a commonality in the approach

More tomorrow. Thanks again!

src/Entity/MetadataDisplayEntity.php
```diff
@@ -441,27 +442,47 @@ public function getTwigVariablesUsed() {
...
    *   A list of used $variables by this template.
    */
-  private function getTwigVariableNames(ModuleNode|Node|BodyNode $nodes): array {
+  private function getTwigVariableNames(ModuleNode|Node|BodyNode $nodes, $all_variables): array {
     $variables = [];
     foreach ($nodes as $node) {
-      if ($node instanceof \Twig\Node\Expression\NameExpression) {
-        $name = $node->getAttribute('name');
-        $variables[$name] = $name;
+      $lineno = [$node->getTemplateLine()];
```
DiegoPino, line 448: nice

src/Entity/MetadataDisplayEntity.php
```diff
@@ -441,27 +442,47 @@ public function getTwigVariablesUsed() {
...
+      $variable_key = '';
+      $parent_path = '';
+      if ($node instanceof \Twig\Node\Expression\NameExpression && !$nodes instanceof \Twig\Node\Expression\Unary\NotUnary) {
```
DiegoPino, line 451: This can be an import

src/Form/MetadataDisplayForm.php
```diff
@@ -174,6 +174,32 @@ public function save(array $form, FormStateInterface $form_state) {
     return $status;
   }
 
+  public static function flattenKeys($array, $recursive_key = '', $recursive_type = '') {
```
DiegoPino, line 177: This function requires typed attributes for the variables

aksm:
Thanks for the feedback @DiegoPino. Re: `::arrayToFlatCommonkeys`, I could reuse and adjust the function here, but it would exclude empty arrays, e.g.

```json
{
  "ismemberof": [],
{
```
If the above is evaluated in the template, the report would list that key as being used, but reusing would ignore it and report that it isn't used. If that's ok, I'd be glad to refactor.

src/Entity/MetadataDisplayEntity.php
```diff
@@ -415,9 +415,9 @@ public function renderNative(array $context) {
   /**
    * {@inheritdoc}
    */
-  public function getTwigVariablesUsed() {
+  public function getTwigVariablesUsed(array $twigtemplate) {
```
DiegoPino, line 418: @aksm I believe this needs a different method. Let's discuss this tomorrow. Overriding the current one makes doing the opposite (what was done before) obscure as hell

DiegoPino (changes requested):

src/Entity/MetadataDisplayEntity.php
```diff
@@ -441,27 +446,47 @@ public function getTwigVariablesUsed() {
    * @return array
    *   A list of used $variables by this template.
    */
-  private function getTwigVariableNames(ModuleNode|Node|BodyNode $nodes): array {
+  private function getTwigVariableNames(ModuleNode|Node|BodyNode $nodes, $all_variables): array {
```
DiegoPino, line 449: @aksm what is $all_variables here? Type? default? I don't see an explanation on the PHP DOC

src/Form/MetadataDisplayForm.php
```diff
@@ -109,7 +109,7 @@ public function buildForm(array $form, FormStateInterface $form_state) {
     $form['preview']['render_native'] = [
       '#type' => 'checkbox',
       '#defaut_value' => FALSE,
-      '#title' => 'Show Preview using native Output Format (e.g HTML)',
+      '#title' => 'Show Preview using native Output Format (e.g HTML) and output errors if there are any.',
```
DiegoPino, line 112: Can you maybe move this to a description?
If errors are found Preview will fail. To checkboxes show "#descrition?"

src/Form/MetadataDisplayForm.php
```diff
@@ -109,7 +109,7 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
       '#states' => [
         'visible' => ['input[name="ado_context_preview"' => ['filled' => true]],
```
DiegoPino, line 114: I think this will revert the fix you already did to #states? Hope note. Please check

src/Form/MetadataDisplayForm.php
```diff
@@ -174,17 +174,41 @@ public function save(array $form, FormStateInterface $form_state) {
     return $status;
   }
 
+  public static function flattenKeys(array $array, string $recursive_key = '', string $recursive_type = '') {
+    $return = [];
+    if (is_array($array)) {
```
DiegoPino, line 180: This is unneeded, the Function signature will not let you pass a non array

src/Form/MetadataDisplayForm.php
```diff
@@ -254,38 +284,92 @@ public static function ajaxPreview($form, FormStateInterface $form_state) {
...
+              if (array_key_exists($used_var_part,$data_json)) {
+                $data_json[$used_var_part]['used'] = 'Used';
+                $data_json[$used_var_part]['line'] = $used_var_line;
+
+              }
+
+            }
+          }
+          else {
+            if (!empty($used_var_parent_path) && str_starts_with($used_var_parent_path, 'data.') && array_key_exists($used_var_parent_path, $data_json)) {
```
DiegoPino, line 327: Suggestion . In general doing `??` for array key existence is faster. Also  isset($used_var_parent_path[$data_json])  is faster

src/Form/MetadataDisplayForm.php
```diff
@@ -397,25 +481,31 @@ public static function ajaxPreview($form, FormStateInterface $form_state) {
...
-        ];
-        $output['json_unused'] = [
-          '#type' => 'details',
-          '#open' => FALSE,
-          '#title' => 'Unused JSON keys',
-          'render' => [
-            'table' => $json_table
-          ],
-        ];
+        if (!empty($json_table)) {
```
DiegoPino, line 487: Don't understand this condition here. Why are you checking for $json_table? and why double $output['json_unused']. Maybe I am missing something?

DiegoPino (changes requested):
Small changes + some questions

src/Entity/MetadataDisplayEntity.php
aksm: @DiegoPino Sorry, forgot about the PHP DOC. Updated via 4fb2803ee2541a7410c2b9808e40ffee28cb91e0 and 790b4ad4d4602ce844db28f4009b07718978f834. Does that make sense?

src/Form/MetadataDisplayForm.php
aksm: Whoops, forgot to rebase. Just fixed now.

src/Form/MetadataDisplayForm.php
aksm: Hmm, I guess this doesn't make sense in the Exception since it means the try didn't work. Will remove.

src/Form/MetadataDisplayForm.php
aksm: Resolved via b80eb4ae69f1b7b8f6b8723a6a4c0eeeffa42de5.

src/Form/MetadataDisplayForm.php
aksm: Resolved via edad28026d576acb959a73f15fdca4c20d7110e4.

src/Form/MetadataDisplayForm.php
aksm: Resolved via 029b2caa68c59d9a55de25fb7547ff8c32110b20.

src/Form/MetadataDisplayForm.php
aksm: @DiegoPino Do you mean have the description only show if the option is checked? If so I couldn't figure it how to do that in time, but I moved it to the description in any case. It seems like it needs to be updated in `ami.module`? Not exactly sure why yet, but the element there seems to override this one.

DiegoPino: @aksm what I mean is to use the #description key in the render array

DiegoPino:
@aksm giving this a try now. Will report with any findings after-noon. Thanks for addressing my review

DiegoPino:
@aksm did a review and there is still work to do here:

 Few important notes:

- There is too much data shown. After checking thoroughly having each internal key of a complex JSON data structure shown as _not being used_ helps a user building a twig template not much. I believe (even if you actually do the current calculation) that, given that the choice of iterating deeper into a structure, e.g as:image is a programatic choice, IF the user uses a single key inside a top JSON key, that JSON key should be marked as **used**. Not all JSON data is there to be used/shown. basically things like  `data.as:image.urn:uuid:5efb8aeb-6c07-42ef-ab8c-b3628dcd7889.url ` marked (also not an array) as not being used are not helpful.

- The Type is not accurate at all. e.g 
  `data.geographic_location.city | array` is a string. Not an array. And this applies to a lot of the keys being displayed.

- Local variables are being parsed incorrectly. 
e.g 

image

See. `item` is a local variables used inside two loops. It is shown as being used only (referenced variable) by one of them. What we need here is:
- Use this current logic to fill up the incorrect reporting of used/not used on the the `JSON keys` part. e.g
On the `JSON keys` section for a given template, data.create_lod.name_label is shows as being used (again also not an array).
image
But then down here, you actually show that `data_creator` is in fact referencing `data.create_lod` thus data_creator.name_label (as being used) is == `data.create_lod.name_label`
image

Conclussion:
- Remove from that table "other variables" that are temporary/internal to loops, etc that is actually a reference to a JSON key and use that logic to report either correctly that the original property is being used, or simply to not count a top element/second level element as "not being used"
- Make sure you either calculate the type correctly or simply we don't used it at all. Wrong info there does not help and the original JSON data allows a user already to see that
- Make the list more concrete. What a user needs to see (we talked this in our team call) is what they are missing at a more general/easy reference level. 

Note: if you check e.g the code I wrote here https://github.com/esmero/strawberryfield/blob/69c1c4e7e44572a548ba6374731f4db66976f295/src/Plugin/DataType/StrawberryKeysFromJson.php#L80
 And in specific here
https://github.com/esmero/strawberryfield/blob/69c1c4e7e44572a548ba6374731f4db66976f295/src/Tools/StrawberryfieldJsonHelper.php#L253

Any property of an array that might end being a very variable one-of (e.g a UUID) is removed.  Not sure if that is helpful for you, but not having every as:document.aspecificUUID and its flvexif, etc shown makes more sense. Right now the code can not track if e.g dr:mimetype -> through an iteration was used and thus mark the as:document.aspecificUUID.dr:mimetype as used or not.

I can meet with you tomorrow to go through this, thanks

DiegoPino (changes requested):
@aksm left a long review as a comment. The code needs further work before merging.

aksm:
@DiegoPino, sorry for the delayed response, but my brain was a little too fried to comprehend all of your thorough notes. I've fixed the types. As for the more complex part, that was the issue I was trying to describe in our initial team call, but I didn't communicate it clearly. I'll retrace my steps and see if I can find an easier way through. Will ping you if I get stuck. Thanks!

DiegoPino:
@aksm ok, let's catch up on this on Monday

src/Entity/MetadataDisplayEntity.php
```diff
@@ -438,30 +443,71 @@ public function getTwigVariablesUsed() {
...
+        $variable_key = $node->getAttribute('value');
+      }
+      elseif ($node instanceof GetAttrExpression) {
+        $variable_names = $this->getTwigVariableNames($node, array_replace_recursive($all_variables, $variables), $set_var, $set_source);
+        $variable_names_flat = [];
+        $prev_name = null;
+        foreach ($variable_names as $name) {
+          $name_check = $prev_name ? $prev_name['path'] . '.' . $name['path']:'';
+          if (($name['path'] == $set_var) && !($set_source == $name_check)) {
+            array_push($variable_names_flat, $set_source);
```
DiegoPino, line 486: `array_push($variable_names_flat, $set_source)` here being $variable_names_flat[] = $set_source ?

src/Entity/MetadataDisplayEntity.php
```diff
@@ -438,30 +443,71 @@ public function getTwigVariablesUsed() {
...
+      }
+      if (!empty($variable_key) && is_string($variable_key)) {
+        $variables[$variable_key]['path'] = $variable_key;
+        if(isset($all_variables[$variable_key]['line'])) {
+          $variables[$variable_key]['line']     = array_unique(array_merge($all_variables[$variable_key]['line'], $lineno));
+          $all_variables[$variable_key]['line'] = array_unique(array_merge($all_variables[$variable_key]['line'], $lineno));
+        }
+        if(!isset($variables[$variable_key]['line'])) {
+          $variables[$variable_key]['line'] = $lineno;
+        }
```
DiegoPino, line 507: This piece here looks strange. So if only present in a single line you make ['line'] a string, if not it is an array. Why not make it an array always? That is more consistent when parsing it on the caller. Also 
```
$variables[$variable_key]['line']     = array_unique(array_merge($all_variables[$variable_key]['line'], $lineno));
           $all_variables[$variable_key]['line'] = array_unique(array_merge($all_variables[$variable_key]['line'], $lineno));
           ````
Is doing the same exactly twice.  I don't see you use $all_variables at all afterwards not return it, but if you need it 
$variables[$variable_key]['line'] = $all_variables[$variable_key]['line'] = array_unique(array_merge($all_variables[$variable_key]['line'], $lineno));

File: src/Form/MetadataDisplayForm.php [hunk 13]
```diff
@@ -174,17 +175,54 @@ public function save(array $form, FormStateInterface $form_state) {
...
+    ++$array_depth;
+    // The below is copied from StrawberryfieldJsonHelper::arrayToFlatJsonPropertypaths to deal with arrays,
+    // but maybe for the next iteration, it would be better to add options to
+    // that existing function and import here.
+    // Blacklist paths. Strip the last dot in case this was called recursively.
+    if (!empty($excludepaths) && in_array(rtrim($recursive_key,'.'), $excludepaths)) {
+      return $flat;
+    }
+    foreach($array as $key=>$value) {
+      if(filter_var($key, FILTER_VALIDATE_URL) || StrawberryfieldJsonHelper::validateURN($key)) {
```
DiegoPino, line 190: Was `StrawberryfieldJsonHelper` included on the header as an USE statement? I don't see it

src/Form/MetadataDisplayForm.php
```diff
@@ -254,38 +296,77 @@ public static function ajaxPreview($form, FormStateInterface $form_state) {
...
-      ];
 
-      // Try to Ensure we're using the twig from user's input instead of the entity's
-      // default.
       try {
+        // Try to Ensure we're using the twig from user's input instead of the entity's
+        // default.
         $input = $form_state->getUserInput();
-        $entity->set('twig', $input['twig'][0], FALSE);
+        $preview_template = $input['twig'];
```
DiegoPino, line 304: Why the extra $preview_template = $input['twig']; and then $preview_template[0] ?

DiegoPino (commented):
Hi, I left a few comments. Won't mark as request changes bc many are questions. Thanks @aksm !

src/Entity/MetadataDisplayEntity.php
```diff
@@ -438,30 +443,71 @@ public function getTwigVariablesUsed() {
...
         }
+        $variable_key = implode('.', $variable_names_flat);
       }
-      elseif ($node instanceof \Twig\Node\Node) {
-        $variables += $this->getTwigVariableNames($node);
+      elseif ($node instanceof Node) {
+        $add_variables = $this->getTwigVariableNames($node, array_replace_recursive($all_variables, $variables), $set_var, $set_source);
+        $variables = array_replace_recursive($variables, $add_variables);
+      }
+      if (!empty($variable_key) && is_string($variable_key)) {
```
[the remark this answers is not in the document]
aksm: Addressed via 29ec509ce49fd890a4fe691558dbb8d7bc72bf78.

src/Entity/MetadataDisplayEntity.php
aksm: >This piece here looks strange. So if only present in a single line you make ['line'] a string, if not it is an array. Why not make it an array always? That is more consistent when parsing it on the caller.

`$lineno = [$node->getTemplateLine()];` above it makes it always an array, no? Did I miss something?

aksm: Re: the `$all_variables`, will update to your suggestion for now (didn't know you could chain assignments, cool!) and address the larger recursion issue in the below comment.

src/Form/MetadataDisplayForm.php
```diff
@@ -174,17 +175,54 @@ public function save(array $form, FormStateInterface $form_state) {
...
+    // Blacklist paths. Strip the last dot in case this was called recursively.
+    if (!empty($excludepaths) && in_array(rtrim($recursive_key,'.'), $excludepaths)) {
+      return $flat;
+    }
+    foreach($array as $key=>$value) {
+      if(filter_var($key, FILTER_VALIDATE_URL) || StrawberryfieldJsonHelper::validateURN($key)) {
+        $key = "*";
+      } elseif (is_integer($key)) {
+        $key = '[*]';
+        //@TODO research implications of $.field[*] versus $.field.[*]
```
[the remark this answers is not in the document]
aksm: Addressed via 35ea752f627241880b7b2a0961cd287136ced3a5.

src/Form/MetadataDisplayForm.php
aksm: https://github.com/esmero/format_strawberryfield/blob/2efe80497d9a336a4ef79846fa2bb1a52a7d4b27/src/Form/MetadataDisplayForm.php#L14

Did I miss an additional step?

src/Form/MetadataDisplayForm.php
aksm: I think I had used that variable to try something else at one point and forgot to change it back. Will fix.

aksm: Addressed via 11c8f2f2b3e7496807fa06488e2eda109911a408.

src/Form/MetadataDisplayForm.php
DiegoPino: You good. Strange the Code preview/diff did not show it. Marked as resolved

src/Entity/MetadataDisplayEntity.php
```diff
@@ -438,30 +443,71 @@ public function getTwigVariablesUsed() {
...
+            array_push($variable_names_flat, $name['path']);
+          }
+          $prev_name = $name;
         }
+        $variable_key = implode('.', $variable_names_flat);
       }
-      elseif ($node instanceof \Twig\Node\Node) {
-        $variables += $this->getTwigVariableNames($node);
+      elseif ($node instanceof Node) {
+        $add_variables = $this->getTwigVariableNames($node, array_replace_recursive($all_variables, $variables), $set_var, $set_source);
```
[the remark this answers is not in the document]
DiegoPino: Not a blocker but the comments/inline explanations of what is happening are important. So please add those (like everywhere you do some complex array gymnastics)

DiegoPino:
@aksm once you add inline comments of what is coming in/out and mostly where you are doing complex array merges I will test again with data and hopefully merge. Thanks

src/Entity/MetadataDisplayEntity.php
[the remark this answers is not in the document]
aksm: Added comments in be20028a3bc6cebb459dfbaa7a3421e1e8f5105c. @DiegoPino you were right about the unnecessary double assignment above. Will return to this later to trim it down more. Thanks!

aksm:
@DiegoPino in addition to adding the comments, I refactored to account for `[*]` and `*` consistently with e599e94ecdc0fa1643716dbc618feb0b4716351b. Hopefully that didn't add any additional mistakes.

DiegoPino:
@aksm will do the last test and merge by 3PM. Thanks again for addressing my review

DiegoPino:
@aksm will approve and merge now. There is an oddity I want to fix in my other open pull request but I want this merged first. E_USER_DEPRECATED should not be caught as errors, and should be logged for the Server Admin. It is annoying for a metadata person to have to deal with things they can not fix by changing/improving the template itself.

DiegoPino (approved):

Issue #302, opened 2023-05-13
General IIIF Viewer Interaction Improvements


DiegoPino:
# What?

There are some very pressing needs when thinking of building interactive, multi viewer exhibit/displays. One of them related to #292 and #297 but more global is to allow generic JS event driven communication between viewers. Use cases
- Mirador or OSD advance a Canvas/Img. Map contains GeoJSON data (img, Points, layers) related to any of these. And the inverse, The map focuses on a particular Point that is bound to an ADO, Mirador/OSD/Bookreader etc could jump/even refine the Manifest based on this interactions
- HTML should contain on every IIIF enabled Viewer a data argument (we can reuse the data-iiif-manifest we already have In some). This allows other JS driven plugins be aware of what is being displayed (independently of the Viewer) is driven by IIIF
- The leaflet Map (both single Viewer based on geoJSON and/or the View driven multi point one with facets) could support IIIF Manifests directly and load (via jack reeds JS Tile Plugin) images
- Annotations on OSD (our building/writing tool) needs to allow tagging via LoD, either from external Sources or constraint to current metadata but also (why not) on Extracted OCR. Basically MiniOCR is already annotation ready, just needs some mangling

That is for now. I have big plans for this. Why this is a focus? Archipelago is already absurdly feature full. Let's do fun/interesting/interactive things with our data.

DiegoPino:
Partially solved via https://github.com/esmero/format_strawberryfield/commit/38e2e5eb434911f1839a4fa6c7fb8634a3a05624

PR #303, opened 2023-05-13
ISSUE-302: A Saga of JS driven IIIF Interactions

DiegoPino:
See #302. Starting here with Mirador and moving into each IIIF driven/enabled (or able to react) Viewer we provide.

WIP, will keep adding commits and comments and will update this once all is done.


DiegoPino:
@esmero/archipelago-core. Merging this, has been in production for 10 days without issues. There is still work todo. e.g add the same override options via JSON to Mirador, make OSD react to Events, etc. But this was becoming too large already

Issue #322, opened 2023-06-15
Parse Metadata Display for Views/embed use to enrich/enhance Cache tags


DiegoPino:
# What?

Twig templates are cached. Views used in Twig templates might have different caching strategy. Which leads sometimes (e.g IIIF CWS manifests) to get cached before the Solr Index has been updated with the Children Info and thus a stuck/empty manifest.

The solution I propose is to:
- Add a Cache dependency parser at the Entity Class level that parses any use of a View/Gets the Views Cache tags and then (yes, again) caches those tags for future use. Why the later? If the cache tags of a View used inside a template are basically "none" we will end parsing the Twig template every time. The idea is we parse them once, cache the tags for the template and add to the cache tags of the cache tags (WOW!) the View as a context. If the views change we have to parse them again, but if not, on actual render we use the parsed tags and add them to the Metadata Display (on the Field Formatter and the Endpoints). This basically allows us to make the best possible caching strategy, always fetch fresh data but also reduce cost of processing. If the template has no views, then all works as before.
This is a release task since its becoming more and more pressing to have exact/fresh data

DiegoPino:
Resolved via https://github.com/esmero/format_strawberryfield/commit/815c591c5148e54815f3d344c57a138be7652663

PR #325, opened 2023-06-20
ISSUE-322: Better related entities/configs cache tag calculation for Metadata Displays

DiegoPino:
See #322 

Instead of using complex search api tags for views calculation (Views use its own search api provided cache class so we can not depend on that) we invalidate parent ADOs on Search API Indexed event (after the Solr document was properly indexed). This way we have dependent tags (exact entities/configs) and also time based/on-time parent invalidation. Tested with background AMI set processing where CWS entities would be cached without listing its children (this happens when Solr Index renders the Entity itself and the Children are not indexed yet, leaving a stale/empty IIIF CWS manifest) and once the Children went into Solr the cache was cleared and all worked. Also, related Metadata Displays (e.g CWS children) updates triggered a CWS metadata display and changes in a View did the same

What this does not do: 
- Doing more than 1 level deep relationship calculation (means 2-3 level of views nesting)

Gotchas:
- a call to twig extension drupal_view('view_name', 'display_name', argument) requires two first arguments to be strings/static ones. If you use a variable to provide those then the caching will not be considered/resolved. This is because we can not render the template/resolve the variable values without initializing the actual Twig Sandbox environment (I can compile, but executing will not work outside that scope). 

@alliomeria this will solve once and for all stale templates and missing listings
@aksm can you do a quick sanity check of my code?

NOTE: because time is a luxury I ended adding in this pull also something related to #324. Basically the story is:
Layout builder (when properly enabled... which funny enough is not just "enabling"... requires an update hook to run that modifies entity view configs, but when provided by our deployment get never modified...) alters view mode (and many other) Routes by adding not only more defaults but also requiring a "service" to be passed to the Form Builder it replaces (the normal View Mode edit form). That extra argument is loaded by the actual route and injected into the Form Class... and we had no way of knowing that. So we have to alter the route/pass arguments around to make sure that, under its presence, we fulfill the needs. If not you will see when pressing on "Active Display Settings" this:

```Error: Call to a member function getLayoutBuilderUrl() on null in Drupal\layout_builder\Form\LayoutBuilderEntityViewDisplayForm->form() (line 66 of core/modules/layout_builder/src/Form/LayoutBuilderEntityViewDisplayForm.php).
```
Being null here the missing Form Function argument...

(Hint: fixed and super well working now)


src/Entity/MetadataDisplayEntity.php
```diff
@@ -138,6 +142,10 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
 
   // Implements methods defined by EntityChangedInterface.
   use EntityChangedTrait;
+  use UseCacheBackendTrait;
+  use LoggerChannelTrait;
+
+  CONST CACHE_TAG_ID = 'format_strawberry:metadata_display_related_tags:';
```
aksm, line 148: @DiegoPino Should this be:

```suggestion
  CONST CACHE_TAG_ID = 'format_strawberryfield:metadata_display_related_tags:';
```

for consistency? Or was it deliberately named that way? Does it make any difference?

DiegoPino: @aksm I kept the prefix shorter bc the view mode resolver and the embargo are using already that prefix. And even if your suggestion is better naming, I would have to change all the other tags and generate a hook update to recreate them under a new Cache bin on update. Feels right now a bit too much.

aksm: Ok, that makes sense. Thanks.

aksm (approved):

DiegoPino:
Thanks @aksm

Issue #326, opened 2023-06-21
Modify (optional) our advanced search behavior to closer (not exact) match community needs


DiegoPino:
and @alliomeria's one too 

# What

We discussed today some needs on Advanced search with @alliomeria and conclusions are:
- add more/remove should have an option for not triggering a search (this is hard since they need to rebuild the form and thus, once Ajax is in place, return a view, this is the Drupal way). I will try to mimic how the "required" filter option works or intercept the query. Issue I see is IF there was already a submission (means I searched for "pumpkin" and got results) and I press "add one", add "Dog" I need to still keep the old results where "pumpkin" was searched. OR I could intercept the Ajax response and simply NOT return any changes, thus the actual View would be still changed and thus the facets would not match... hard
- When using "required" option on the advanced search filter, we can actually start with an empty view but we have no control on the message that displays (No results found). the idea here is that if no input has been given we either use the Message provided by the Advanced Tab, Exposed Input, Input required special form 
![image](https://github.com/esmero/format_strawberryfield/assets/6946023/917636eb-1adb-4ead-9dec-c31ef31d2448) or none, but never "No results found"
- Pressing enter on the form should not trigger add one/remove one. Should always submit the search.
- Selection of Field names should allow a "rewrite". That requires a new pattern/field with replacement options.

So far this is what I have. This is a lot of code but sadly for me needed for the release.

DiegoPino:
😢

DiegoPino:
Solved via https://github.com/esmero/format_strawberryfield/commit/715eb8f1e827a246851afaea960e0f8387d38682

PR #327, opened 2023-06-21
ISSUE-326: Advanced search community needs

DiegoPino:
See #326 

First one:
Submit on enter (not add one/remove one on Enter)

So much code for such a simple thing @alliomeria . Stuff is: Browsers will always pick the first submit found, so rendering order leads. If i put search first then other stuff like adding a + button by each one will still fail (future needs). Also Drupal will make all  buttons input type=submit (if we want to pass stuff via form state) so i have to alter the form, intercept the Enter and trigger the only and only one (staring at the moon) button i care for (i care of other things, like plants and some people and trees and moonshine)

Second: no more stuck keywords on remove one, add one. I changed the whole logic, drupal is so buggy. Now Add more/remove one act on pure JS, change values and then autosubmit (via drupal). Which in theory means in the near future I can have add more/remove without any submits at all.. (not today). Also no more ghost "previous" values appearing

Third: Fields to be searched can have their Labels rewritten (don't like Rendered HTML? you can name them `Pikachu field` and done

Fourth: You can define a min of initial fields. Don't like a single one? Start with 2!

Need to show you this so you can find more bugs tomorrow, but I feel release is ready to roll if all goes well/passes your checks

### DIFF
--- a/modules/format_strawberryfield_facets/src/Plugin/facets_summary/processor/LastActiveFacetsProcessor.php
+++ b/modules/format_strawberryfield_facets/src/Plugin/facets_summary/processor/LastActiveFacetsProcessor.php
@@ -133,7 +133,7 @@ class LastActiveFacetsProcessor extends ProcessorPluginBase implements BuildProc
         $exposed_input = $view->getExposedInput();
         $view->getRequest()->getRequestUri();
         $keys_to_filter = [];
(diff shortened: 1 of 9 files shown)


alliomeria:
🚀 🌟 Woohoo! So many great advanced search toolings @DiegoPino! 🧰 🤓

PR #344, opened 2023-07-26
ISSUE-343: Metadata Preview error display

### LINKED ISSUE #343: Metadata Preview error display
# What?
Related to #338. Currently both Twig syntax errors and libxml errors are displayed to the user inconsistently, and the latter appears as part of the HTML output body.

One possible solution is to provide a function that generates a container with an error status message at the top of the preview window.

aksm:
Resolves #343. Adds a static function that inserts a container with an error status at the top of the Metadata Preview modal if there is an error message and replaces the error messages that were being displayed in the details tabs.

### DIFF
--- a/src/Form/MetadataDisplayForm.php
+++ b/src/Form/MetadataDisplayForm.php
@@ -234,26 +234,24 @@ class MetadataDisplayForm extends ContentEntityForm {
   }
 
   /**
-   * Takes an output array and an error message and returns
-   * the output with a status message container at the top.
+   * Takes an error message and returns
+   * the status message container.
    */
-  public static function buildAjaxPreviewError(array $output, string $message) {
-    if(!empty($message)) {
-      $output['preview_error'] = [
-        '#type' => 'container',
-        '#weight' => -1000,
-        '#theme' => 'status_messages',
-        '#message_list' => [
-          'error' => [
-            t($message),
-          ],
+  public static function buildAjaxPreviewError(string $message) {
+    $preview_error = [
+      '#type' => 'container',
+      '#weight' => -1000,
+      '#theme' => 'status_messages',
+      '#message_list' => [
+        'error' => [
+          t($message),
         ],
-        '#status_headings' => [
-          'error' => t('Error message'),
-        ],
-      ];
-    }
-    return $output;
+      ],
+      '#status_headings' => [
+        'error' => t('Error message'),
+      ],
+    ];
+    return $preview_error;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


DiegoPino:
@aksm before reviewing this, can you please also update AMI one? Or we don't need the change there? I prefer to get that ISSUE solved in both places where needed. Thanks!

aksm:
> @aksm before reviewing this, can you please also update AMI one? Or we don't need the change there? I prefer to get that ISSUE solved in both places where needed. Thanks!

@DiegoPino Taking a look now.

src/Form/MetadataDisplayForm.php
```diff
@@ -389,17 +412,12 @@ public static function ajaxPreview($form, FormStateInterface $form_state) {
...
-              '#attributes' => [
-                'class' => ['error'],
-              ],
-            ],
             'render' => [
                 '#markup' => $render,
             ],
           ];
         }
+	$output = static::buildAjaxPreviewError($output, $message);
```
DiegoPino, line 420: @aksm sorry for being picky here, but it makes more sense to to leave the logic of "is there a message to be shown" to the caller than to the actual buildAjaxPreviewError method. means the  if(!empty($message)) { $output = static::buildAjaxPreviewError($output, $message);
}

Also, not sure if passing the whole rendered (big variable) around to another method to add a key and then back (these are not Objects so you are actually passing a huge value/memory) is the best approach. 

What if static::buildAjaxPreviewError just generates the piece. And then you add to the existing $output that key?

Does this make sense to you?

DiegoPino (changes requested):
Move the conditional logic (if the message is empty) out of the static method. Also don't pass the whole/big $output array, in other words make the  ::buildAjaxPreviewError to just that, generate a render array that you add in the caller side of things. Even if that adds more code (a few extra lines) it allows you to basically reuse  public static function buildAjaxPreviewError under other keys, in other hierarchies and avoids passing (not by reference, so it is an actual copy in memory) a huge piece of data around

src/Form/MetadataDisplayForm.php
aksm: All makes sense and good to know. Will make the changes and ping you. Thanks!

src/Form/MetadataDisplayForm.php
```diff
@@ -233,6 +233,27 @@ public static function addPropertyPath(string $property_path) {
     ];
   }
 
+  /**
+   * Takes an error message and returns
+   * the status message container.
+   */
+  public static function buildAjaxPreviewError(string $message) {
```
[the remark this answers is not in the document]
aksm: Resolved via e2684a51b1a1892e72ff52eb5902fd43d7a04978.

DiegoPino (approved):
Great. Now we can merge

PR #347, opened 2023-07-27
ISSUE-338: Custom error handler needs to filter out unhelpful errors/deprecations

### LINKED ISSUE #338: Custom error handler needs to filter out unhelpful errors/deprecations
# What?
Related to #320. The custom error handler was throwing exceptions on unrelated errors and deprecations, which were being reported to an end user (who wouldn't be able to do anything about it) editing metadata templates. We should limit these to only what's useful for editing/creating templates.

aksm:
Resolves #338

### DIFF
--- a/src/Form/MetadataDisplayForm.php
+++ b/src/Form/MetadataDisplayForm.php
@@ -286,6 +286,10 @@ class MetadataDisplayForm extends ContentEntityForm {
       $mimetype = !empty($mimetype) ? $mimetype[0]['value'] : 'text/html';
       $show_render_native = $form_state->getValue('render_native');
 
+      if ($show_render_native) {
+        set_error_handler('_format_strawberryfield_metadata_preview_error_handler');
+      }
+
       $sbf_fields = \Drupal::service('strawberryfield.utility')
         ->bearsStrawberryfield($preview_node);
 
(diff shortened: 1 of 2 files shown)


DiegoPino (commented):
@aksm small/tiny change. Can you please update the  *   Throw ErrorException for all errors passed from trigger_error(). It now only throws errors on USER warning and USER error

src/Form/MetadataDisplayForm.php
```diff
@@ -437,6 +441,10 @@ public static function ajaxPreview($form, FormStateInterface $form_state) {
 
 	}
       }
+      if ($show_render_native) {
+        restore_error_handler();
+        restore_exception_handler();
```
DiegoPino, line 446: Just a question, where do we replace the exception handler? Does it require restoring? I see the error handler. Or are both needed?

aksm: Started looking into this this morning, but didn't get back to it. Will take a look again tomorrow.

aksm: Hi @DiegoPino, I originally added this because I thought I'd seen a Stack Overflow answer at one point that mentioned needing to use the restore_exception_handler even if it wasn't set for some reason, but I can't remember what it was about and can't find it anymore so I've removed it with c39da32.

DiegoPino (approved):
This looks good now. Thanks

Issue #348, opened 2023-07-31
Add Node Entity Caster for Views Driven %node paths


DiegoPino:
# What?

Currently it is impossible to add a Tab driven by a View with a `/node/%node/somepath` page path and using `access control` at the same time. The reason is: the way parameter casting/route enriching in Drupal (for the last 8 years reported) keeps the %node as a string instead of a `NodeInterface`, as it should ,bc no parameter upcasting survives the `access control` logic at the route level of a view.

See https://www.drupal.org/project/drupal/issues/2528166 (and see the last patch, it is basically moving parameter bag setting from one place to another). We need to have `NodeInterface` upcasting on our active display route because of our need to enrich that Route with the View Mode and the Bundle so we can load the View Mode Settings Form.

The solution is not that complex. basically identify if a `%node` is being called by a View Generated Route and intercept in a classy/elegant way here: https://github.com/esmero/format_strawberryfield/blob/974bb2f4d984f949c93a016963e0517ef989f200/src/Plugin/Menu/LocalTask/ViewModeLocalTask.php#L57-L69 by adding an extra check, loading the Node and then adding the missing arguments.

Tested an initial (not classy, not elegant) implementation and it solves the issue.

Funny enough,  I also found this sloppy code here
https://git.drupalcode.org/project/bootstrap_barrio/-/blob/5.x/bootstrap_barrio.theme#L595-605

This one fails if one tries to access a Valid Route with a non existing Drupal Node ID. Gosh people. So easy as not trying to access a Property on a NULL.

Anyhow. This will be fixed tomorrow so I can keep site building for 1.1.0 (1.1.1 by now)

DiegoPino:
Resolved (until Drupal 11.x) via https://github.com/esmero/format_strawberryfield/commit/6860c1837bd67aa71153a63edeac64c483570ccb

PR #349, opened 2023-08-01
ISSUE-348: Adds a pseudo Node "Upcaster" for Routed Views in the presence of "Active Display Settings" tab

DiegoPino:
See #348 

This tries to get around the exception that happens when a View uses %node as a parameter, but does not upcast it the value to NodeInterface. It (of course) does this without too much overhead, which means it does not processes the View  Entity Itself and we simply assume a `view_id` + `node `argument are reasons enough to enrich the Route. It is important to node that really only place where this can happen (so pretty edge case) is when a View's path is node/%node given that (fixed in code) that is where our Route (Active Display settings) will call this class. Basically when Node level Tabs (actions) and View level Actions depend on each other/live side by side.


DiegoPino:
Thanks, addressed the wording too. Thanks for catching that.

Issue #306, opened 2023-05-24
Encapsulate Twig variable functionality to reuse in AMI


aksm:
# What?
In order to facilitate esmero/ami#165, the following should be encapsulated into a function that can be reused there:

https://github.com/esmero/format_strawberryfield/blob/afdac188da8e824d907cf6f0b9b215f3ae43629a/src/Form/MetadataDisplayForm.php#L325-L371

Issue #318, opened 2023-06-05
Twig template label is null when adding templates


aksm:
# What?
While investigating/testing #315, I got this error:

> TypeError: Twig\Source::__construct(): Argument #2 ($name) must be of type string, null given, called in /var/www/html/web/modules/contrib/format_strawberryfield/src/Entity/MetadataDisplayEntity.php on line 423 in Twig\Source->__construct() (line 30 of /var/www/html/vendor/twig/twig/src/Source.php)

The label is null here for newly added templates:
https://github.com/esmero/format_strawberryfield/blob/74d68f93099ec2c94976797f33dbe86b58e53a40/src/Entity/MetadataDisplayEntity.php#L423

I guess I had never tested `getTwigVariablesUsed()` while adding a new template? Or something else has changed? Will investigate further.

DiegoPino:
@aksm, sadly there are too many issues in SBF related to this new piece of code. I can not get around deprecation errors rendering but e.g not triggering views on templates and it is not clear to me if Pat's comment is or not valid + plus this one. This piece of archipelago is core. I will try (12AM here) to revert the changes to a previous state in a while. And when I come back we will have to re-work the code into archipelago. Won't remove the actual getVariables code from the entity itself but I will remove the usage on the preview. We talked about this with @alliomeria and we agree stability wins here. Thanks

DiegoPino:
resolved via https://github.com/esmero/format_strawberryfield/commit/663b0fa0987738adbc8c451c11d55322dccb75b0

PR #351, opened 2023-08-02
ISSUE-350: Refactor Metadata Preview used variables functionality

### LINKED ISSUE #350: Refactor Metadata Preview used variables functionality
# What?
Following #320, the disabled functionality needs to be refactored to deal with edge cases before it can be reenabled.

aksm:
Resolves #350, #318, and #306.

@DiegoPino I believe this rearrangement bypasses the edge case, but I wasn't able to reproduce those errors so not 100% sure. Even if it does cover those cases, I think this could definitely use further simplification. I'd be glad to open another issue to keep improving on that front later if this looks ok.

### DIFF
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -454,7 +454,7 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
       $source = new Source($twigtemplate, $this->label() ?? '', '');
       $tokens = $this->twigEnvironment()->tokenize($source);
       $nodes = $this->twigEnvironment()->parse($tokens);
-      $used_vars = $this->getTwigVariableNames($nodes,[]);
+      $used_vars = $this->getTwigVariableNames($nodes);
       ksort($used_vars);
       $this->usedTwigVars = $used_vars;
     }
@@ -480,8 +480,9 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
           if (is_array($value) && isset($merged[$key]) && is_array($merged[$key])) {
             $merged[$key] = static::arrayMergeRecursive($merged[$key], $value);
           }
-          else if(is_array($value) && $key == 'line') {
-            $merged[$key] = array_unique($value,SORT_NUMERIC);
+          else if (isset($value['line']) && count($value['line']) > 1) {
+            $value['line'] = array_unique($value['line'], SORT_NUMERIC);
+            $merged[$key] = $value;
           } else {
             $merged[$key] = $value;
           }
@@ -513,20 +514,51 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
     $variables = [];
     foreach ($nodes as $node) {
       $lineno = [$node->getTemplateLine()];
-      $variable_key = $node->hasAttribute('name') ? $node->getAttribute('name'):'';
-      if($variable_key == 'data') {
-        $variable_key_suffix = $nodes->getNode('attribute')->getAttribute('value');
-        $variable_key = $variable_key . '.' . $variable_key_suffix;
+      $variable_key = $node->hasAttribute('name')
+                      && $node->getAttribute('name') == 'data'
+        ? $node->getAttribute('name')
+        : '';
+      if ($node->hasAttribute('always_defined')) {
+        if ($node->getAttribute('always_defined')
+            && $nodes->hasNode('seq')) {
+          $seq = $nodes->getNode('seq');
+          if ($seq->hasNode('node')
+              && $seq->getNode('node')->hasAttribute('name')
+              && $seq->getNode('node')->getAttribute('name') == 'data'
+          ) {
+            $seq_name = $seq->getNode('node')->getAttribute('name');
+          }
+          elseif ($seq->hasAttribute('name')
+                  && $seq->getAttribute('name') == 'data') {
+            $seq_name = $seq->getAttribute('name');
+          }
+          if (!empty($seq_name)) {
+            $seq_value = $seq->hasNode('attribute')
+              ? $seq->getNode('attribute')->getAttribute('value')
+              : '';
+            $variable_key = empty($seq_value) ? $seq_name : $seq_name . '.' . $seq_value;
+          }
+        }
       }
-      if(!empty($variable_key)) {
+      elseif ($node->hasAttribute('value')
+              && $nodes->hasNode('node')
+              && $nodes->getNode('node')->hasAttribute('name')
+              && $nodes->getNode('node')->getAttribute('name') == 'data'
+      ) {
+        $variable_key = $nodes->getNode('node')->getAttribute('name')
+                        . '.' . $node->getAttribute('value');
+      }
+      if (!empty($variable_key && str_starts_with($variable_key, 'data.'))) {
         $variables[$variable_key]['path'] = $variable_key;
+        $variables[$variable_key]['line'] = isset($variables[$variable_key]['line'])
+          ? array_merge($variables[$variable_key]['line'], $lineno)
+          : $lineno;
       }
       if ($node instanceof Node) {
-        if (!empty($variable_key)) {
-          $variables[$variable_key]['line']  = isset($variables[$variable_key]['line']) ? array_merge($variables[$variable_key]['line'], $lineno) : $lineno;
-        }
         $add_variables = $this->getTwigVariableNames($node);
-        $variables = static::arrayMergeRecursive($variables, $add_variables);
+        if (!empty($variables) || !empty($add_variables)) {
+          $variables = static::arrayMergeRecursive($variables, $add_variables);
+        }
       }
     }
     return $variables;
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -491,12 +491,9 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
       $lineno = [$node->getTemplateLine()];
       $variable_key = '';
       $parent_path = '';
-      if ($node instanceof NameExpression && !$nodes instanceof NotUnary) {
-        if (!$node->hasAttribute('name')) {
-          continue;
-        }
+      if ($node->hasAttribute('name')) {
         $variable_key = $node->getAttribute('name');
-        if ($variable_key == '_key') {
+        if ($node instanceof NameExpression && !$nodes instanceof NotUnary && $variable_key == '_key') {
           $seq = $nodes->getNode('seq');
           if ($seq->hasNode('node') && $seq->getNode('node')->hasAttribute('name')) {
             $seq_name = $seq->getNode('node')->getAttribute('name');
@@ -504,15 +501,13 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
           elseif ($seq->hasAttribute('name')) {
             $seq_name = $seq->getAttribute('name');
           }
-          else {
-            continue;
+          if($seq->hasNode('attribute')) {
+            $seq_value = $seq->getNode('attribute')->getAttribute('value');
           }
-          $seq_value = $seq->hasNode('attribute') ? $seq->getNode('attribute')->getAttribute('value') : '';
           $value_target = $nodes->getNode('value_target');
-          if (!$value_target->hasAttribute('name')) {
-            continue;
+          if ($value_target->hasAttribute('name')) {
+            $variable_key = $value_target->getAttribute('name');
           }
-          $variable_key = $value_target->getAttribute('name');
           $parent_path = empty($seq_value) ? $seq_name : $seq_name . '.' . $seq_value;
           $set_var = $variable_key;
           $set_source = $parent_path;
--- a/src/Entity/MetadataDisplayEntity.php
+++ b/src/Entity/MetadataDisplayEntity.php
@@ -506,7 +506,6 @@ class MetadataDisplayEntity extends ContentEntityBase implements MetadataDisplay
         $variable_key = 'data.' . $node->getAttribute('value');
       }
       if (!empty($variable_key)) {
-        $variables[$variable_key]['path'] = $variable_key;
         $variables[$variable_key]['line'] = isset($variables[$variable_key]['line'])
           ? array_unique(array_merge($variables[$variable_key]['line'], $lineno))
           : $lineno;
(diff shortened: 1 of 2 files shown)


DiegoPino:
@aksm thanks. I will do a deeper testing tomorrow and let you know how it goes. More soon!

aksm:
Hi @DiegoPino, I just updated some PHPDocs and resolved #306 with 74c0ba3 so adding this to AMI can happen right after this is merged.

aksm:
@DiegoPino I just pushed a commit with the option to toggle this functionality off.

Also, re: local variables in for loops, as far as I can tell, this doesn't touch them.

DiegoPino:
Great. Will do some testing tonight. Thanks 😊 --
Diego Pino Navarro
Digital Repositories Developer
Metropolitan New York Library Council (METRO)

DiegoPino:
@aksm testing now

DiegoPino (changes requested):
@aksm ok, still issues.

Loaded a IIIF Manifest (v3) with an ADO.

NOTICE: PHP message: TypeError: Drupal\format_strawberryfield\Entity\MetadataDisplayEntity::getTwigVariableNames(): Argument #4 ($set_source) must be of type string, null given, called in /var/www/html/web/modules/contrib/format_strawberryfield/src/Entity/MetadataDisplayEntity.php on line 543 in /var/www/html/web/modules/contrib/format_strawberryfield/src/Entity/MetadataDisplayEntity.php on line 488 #0 /var/www/html/web/modules/contrib/format_strawberryfield/src/Entity/MetadataDisplayEntity.php(543): Drupal\format_strawberryfield\Entity\MetadataDisplayEntity->getTwigVariableNames(Object(Twig\Node\Expression\Binary\RangeBinary), Array, 'pagenumber', NULL)
#1 /var/www/html/web/modules/contrib/format_strawberryfield/src/Entity/MetadataDisplayEntity.php(543): Drupal\format_strawberryfield\Entity\MetadataDisplayEntity->getTwigVariableNames(Object(Twig\Node\ForNode), Array, 'pagenumber', NULL)
#2 /var/www/html/web/modules/contrib/format_strawberryfield/src/Entity/MetadataDisplayEntity.php(543): Drupal\format_strawberryfield\Entity\MetadataDisplayEntity->getTwigVariableNames(Object(Twig\Node\Node), Array, 'document', 'data.as:documen...')

Which means you are not checking if the value passed in the recursive calls to getTwigVariableNames might/not be null..

Also, DCS there are a few (2 I saw) `if(` without a space before.

Give this another run going via the debugger. Maybe the fact that you are depending too much on the Function's default value for arguments makes you be less aware of what you are passing.

DiegoPino:
Hint.  $set_source = $parent_path; You sure $parent_path might not be null?

aksm:
@DiegoPino, ok, sorry I missed that. One other thing I just noticed while debugging this is the following E_USER_DEPRECATED error:

> Using an "if" condition on "for" tag in "IIIF Presentation API 3.0 Manifest" at line 183 is deprecated since Twig 2.10.0, use a "filter" filter or an "if" condition inside the "for" body instead (if your condition depends on a variable updated inside the loop).

Isn't that something we want to show to the user? Should I check these errors for any mention of "Twig"?

DiegoPino:
Question about E_USER_DEPRECATED  you mention, is @aksm: is this avoiding an actual rendering? Means also Search API deprecations (test please with the Search Solr extension we have in archipelago) are permeating/hindering a preview?

DiegoPino:
@aksm also: before I test again. Could you please give this another pass/test with real data/ various templates and extensions?
I'm happy to test Monday noon if that means the code is better tested.

aksm:
Hi @DiegoPino, I've reduced the recursive Twig variable function by a significant amount but had to add a custom implementation of PHP's array_merge_recursive(). I think this could use further refinement, but I hope it's ok for now. Will do a little further testing with more ADOs and templates and then move onto the display behavior.

DiegoPino:
@aksm thanks. let me know when you are done testing and I can give it a pass 😸

aksm:
Hi @DiegoPino, I believe this is ready for review now. Thanks!

DiegoPino:
@aksm thanks. Will review starting at 2PM EST and let you know how this goes!

src/Entity/MetadataDisplayEntity.php
```diff
@@ -473,95 +507,57 @@ public function getTwigVariablesUsed() {
...
-        if (!empty($parent_path)) {
-          $variables[$variable_key]['parent_path'] = $parent_path;
+        $variables[$variable_key]['line'] = isset($variables[$variable_key]['line'])
+          ? array_merge($variables[$variable_key]['line'], $lineno)
+          : $lineno;
+      }
+      if ($node instanceof Node) {
+        $add_variables = $this->getTwigVariableNames($node);
+        if (!empty($variables) || !empty($add_variables)) {
+          $variables = static::arrayMergeRecursive($variables, $add_variables);
```
DiegoPino, line 560: There is excessive recursion here. Basically the arrayMergeRecursive will traverse over and over the same keys (said differently, If the array has N entries, you will traverse N times the last loop, N-1 time the almost last, N-2, etc), when you could really just traverse N times or not at all using https://github.com/esmero/format_strawberryfield/pull/351/files#diff-b5161fcfb4c915fb64182570502fc39348dd9af67772fa46c6e2a7f24f814e3eR553 and passing $variables as a/by reference argument between recursions.

src/Entity/MetadataDisplayEntity.php
```diff
@@ -473,95 +507,57 @@ public function getTwigVariablesUsed() {
...
-        if (!$node->hasAttribute('name')) {
-          continue;
-        }
-        $variable_key = $node->getAttribute('name');
-        if ($variable_key == '_key') {
+      $variable_key = $node->hasAttribute('name')
+                      && $node->getAttribute('name') == 'data'
+        ? $node->getAttribute('name')
+        : '';
+      if ($node->hasAttribute('always_defined')) {
```
DiegoPino, line 521: There is no need to parse anything at all (you can return) if $variable_key is empty (also better approach is to make it NULL instead of ''. That is cheaper and more consistent when checking if.

src/Entity/MetadataDisplayEntity.php
```diff
@@ -473,95 +507,57 @@ public function getTwigVariablesUsed() {
...
         }
       }
-      elseif ($node instanceof ConstantExpression && $nodes instanceof GetAttrExpression) {
-        $variable_key = $node->getAttribute('value');
+      elseif ($node->hasAttribute('value')
+              && $nodes->hasNode('node')
+              && $nodes->getNode('node')->hasAttribute('name')
+              && $nodes->getNode('node')->getAttribute('name') == 'data'
+      ) {
+        $variable_key = $nodes->getNode('node')->getAttribute('name')
```
DiegoPino, line 548: Everywhere here $variable_key (first part) will always start with `data` (because of the conditions). I see how you could save  a lot of IFs by grouping conditionals and also setting some values as fixed strings (e.g data)

DiegoPino (changes requested):
Hi, first pass, there is some performance issues and also some IF/ELSE that are not even needed (you can break before even going there?). If you can add some inline comments explaining what/type of variable parsing is being done by each conditional, that would help me a lot. Like what is a `seq`? When `always_defined` works, etc

aksm:
Hi @DiegoPino, I believe I've addressed your above comments with the most recent commits, and I think I'm better able to explain my reasoning this time. Will do a little more testing and double-checking to be sure.

DiegoPino:
@aksm still reviewing it but need to pull again after https://github.com/esmero/format_strawberryfield/pull/351/commits/9c059efa31cf84f3aea8dcd5ba48444eea73986c. I need another hour, slow today.

src/Form/MetadataDisplayForm.php
```diff
@@ -502,11 +502,11 @@ public static function ajaxPreview($form, FormStateInterface $form_state) {
           $message = $exception->getMessage();
         }
       } finally {
-        if(!empty($message)) {
+        if (!empty($message)) {
           $preview_error = static::buildAjaxPreviewError($message);
           $output['preview_error'] = $preview_error;
         }
-        if (!$show_render_native || ($show_render_native && $mimetype != 'text/html')) {
+        if ($render && (!$show_render_native || ($show_render_native && $mimetype != 'text/html'))) {
```
aksm, line 509: When `$render = $entity->renderNative($context);` (line 430 above) throws an exception, e.g. the template can't be rendered because of bad Twig, this generates a warning because `$render` is now undefined. Additionally, since the `$show_json_table` assignment comes after, it also generates the undefined warning in the logs. Should these variables be initialized before?

@DiegoPino thoughts?

aksm: or maybe set $render to false or null in the catch if not set, or add an isset check for both variables?

aksm: Ok, thought it through a little further, and it seems like it makes the most sense to use isset($render) instead of just $render in the conditionals. Will make another commit shortly.

DiegoPino: yep, that seems like easier. Thanks

DiegoPino: Or you can put another try/catch around $render = $entity->renderNative($context); ?

aksm: > Or you can put another try/catch around $render = $entity->renderNative($context); ?

@DiegoPino Is there an advantage to nesting another try/catch? Seems like the outer one catches the same exception either way?

DiegoPino: Not really. I would require to refactor your try catch into two. The issue is because you have a final/ the catch itself (or the final) will have to do the extra work. Just dismiss. the isset() is  enough

aksm:
Ok @DiegoPino, sorry about that. I've now tested with all output mime types, multiple output options, multiple templates (new and saved), different exceptions, and even AMI set rows.

DiegoPino:
All good. Will have this reviews/tested by tomorrow and we check back on Monday. Have a good weekend

DiegoPino (changes requested):
@aksm still doing testing here. The actual variable processing looks better (DCS still not perfect) but there is a lot of redundancy in how you are processing that table on the output + DCS. I will keep testing for an hour and then request more changes/have a talk about this during our Team call. Hope we can get it in decent shape for a merge today

src/Entity/MetadataDisplayEntity.php
```diff
@@ -473,96 +473,50 @@ public function getTwigVariablesUsed() {
...
-   *   An string to track the referenced JSON key for the above.
+   * @param array $variables
+   *   The variables passed through recursions.
    *
    * @return array
    *   A list of used $variables by this template.
    */
-  private function getTwigVariableNames(ModuleNode|Node|BodyNode $nodes, array $all_variables, string $set_var = '', string $set_source = ''): array {
-    $variables = [];
+  private function getTwigVariableNames(
```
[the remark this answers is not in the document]
aksm: Hi @DiegoPino. I've been keeping the coding standards open and referring to them, but I missed that section while jumping between standard PHP conventions and Drupal-specific ones. Will update.

src/Entity/MetadataDisplayEntity.php
```diff
@@ -473,96 +473,50 @@ public function getTwigVariablesUsed() {
...
-        foreach ($variable_names as $name) {
-          $name_check = $prev_name ? $prev_name['path'] . '.' . $name['path']:'';
-          if (($name['path'] == $set_var) && !($set_source == $name_check)) {
-            $variable_names_flat[] = $set_source;
-          }
-          else {
-            $variable_names_flat[] = $name['path'];
-          }
-          $prev_name = $name;
+      // Parse seq to get the values for for/in loops,
```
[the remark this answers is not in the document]
aksm: Sorry, maybe not the clearest example. I meant getting/checking the "name" attribute for "data". Will update with a better example.

src/Form/MetadataDisplayForm.php
```diff
@@ -287,45 +219,32 @@ public static function buildAjaxPreviewError(string $message) {
...
-      $used_var_path = $used_key;
-      $used_var_line = $used_var['line'];
-      $used_var_exploded = explode('.', $used_var_path);
-      array_push($used_keys, $used_var_path);
-      $wildcard_paths = static::addPropertyPath($used_var_path);
-      if (isset($data_json[$used_var_path])) {
-        $data_json[$used_var_path]['used'] = 'Used';
-        $data_json[$used_var_path]['line'] = $used_var_line;
+    $data_json = [];
+    foreach($jsondata as $key=>$value) {
```
DiegoPino, line 223: DCS > `$key=>$value` should be `$key => $value`

src/Form/MetadataDisplayForm.php
```diff
@@ -287,45 +219,32 @@ public static function buildAjaxPreviewError(string $message) {
...
-        }
-        if (count($used_var_exploded) > 2) {
-          $used_var_parts = array_slice($used_var_exploded,0, 2);
-          $used_var_part = implode('.', $used_var_parts);
-          if (isset($data_json[$used_var_part])) {
-            $data_json[$used_var_part]['used'] = 'Used';
-            $data_json[$used_var_part]['line'] = $used_var_line;
-          }
-        }
+      elseif (is_array($value) && !StrawberryfieldJsonHelper::arrayIsMultiSimple($value)) {
```
DiegoPino, line 229: Did ::arrayIsMultiSimple do the trick? 😄

src/Form/MetadataDisplayForm.php
```diff
@@ -287,45 +219,32 @@ public static function buildAjaxPreviewError(string $message) {
...
+      $data_json[$key]['used'] = !empty($used_lines) ? 'Used' : '';
+      $data_json[$key]['line'] = $used_lines;
+      $value_type = empty($value) && !is_null($value) ? 'empty ' . gettype($value) : gettype($value);
+      $data_json[$key]['type'] = $value_type;
     }
-    $unused_vars = $data_json;
+    ksort($data_json);
 
-    $unused_keys = array_keys($unused_vars);
+    $unused_keys = array_keys($data_json);
```
DiegoPino, line 239: good, no more duplicated json

aksm:
Hi @DiegoPino, aside from a little more cleanup and maybe some comments/notes, I think this is in much better shape. If we need to go deeper into the ADO JSON keys, it'll be much easier to add that later instead of starting complex and simplifying (my biggest mistake here). Sorry for the delay, but this is turning out to be a great learning experience for me.

aksm:
Hi @DiegoPino, I think the changes based on the recent feedback from you and the team are ready for review. I will do some further looking over while I continue with the documentation.

DiegoPino (approved):
Code is simpler now and I feel most concerns are addressed. Merging. Thanks @aksm

Issue #363, opened 2023-09-26
Map formatter has invalid Validations on Config Form initializing


DiegoPino:
# What?

This only becomes apparent when using the formatter directly on a View/Field bc the Form is validated even before is setup failing badly via AJAX

e.g

```
An AJAX HTTP error occurred.
HTTP Result Code: 200
Debugging information follows.
Path: /admin/structure/views/ajax/handler/map_display_via_solr/block_1/field/field_descriptive_metadata
StatusText: OK
ResponseText: TypeError: array_filter(): Argument #1 ($array) must be of type array, null given in array_filter() (line 238 of /var/www/html/web/modules/contrib/format_strawberryfield/src/Plugin/Field/FieldFormatter/StrawberryMapFormatter.php).
```

patdunlavey:
Hi Diego, I wonder if I can piggy-back another small issue on this one since it's still open, and I found it while looking into the view form error? 

If you look [here](https://github.com/esmero/format_strawberryfield/blob/a2447705572dd64d41c6f8336c80def5625be10f/src/Plugin/Field/FieldFormatter/StrawberryMapFormatter.php#L921-L924), we see that it's not possible to display a map shorter than 480 pixels high. However in my case I want to display it with a height of just 200 pixels. Is it actually necessary to impose a minimum height at all? I think a site builder should be trusted to select whatever value suits them, but we could always just set a minimum height of, say, 10, just to ensure that they see something?

DiegoPino:
@patdunlavey will look at it. I think the 480 might be a legacy thing when I wrote that formatter 5 years ago, but there is indeed a min height limit for leaflet (will look at it and make sure I document as such). Thanks

DiegoPino:
@patdunlavey but 1.2 is closed. so the fix you want will go into 1.3.0. Hope that is OK

patdunlavey:
Understood. Thanks!

DiegoPino:
Oh, wait. 1.2 is still open. Will make a dual pull. More tomorrow

DiegoPino:
Hi, yes thanks 😊

patdunlavey:
Ok, closing. New issue for the leaflet settings created [here](https://github.com/esmero/format_strawberryfield/issues/414).

Thanks!

PR #364, opened 2023-09-26
ISSUE-363: Wrong empty defaults break View Ajax settings on Map formatter

DiegoPino:
See #363 

This value can be in fact be null but bc we have to compare (when it has one) against valid sources, so we do an array filter and break hell loose. This should fix it


DiegoPino:
Ok, tested. All looks good!

Issue #368, opened 2023-09-29
We need a new EDTF Twig extension to be able to parse dates for Citations


DiegoPino:
# What?

Ok, let's start by pinging @alliomeria here. The issue is this: currently with the humanizer
If I get `19XX-XX-XX` I will get `1910s` humanized and I need `(1900)` in e.g APA
If I get `1912?-1X-XX` I will get `October 1912 (year is uncertain and approximate) ` and I need `(1912, October)` in APA
As you notice the combinations of the humanizer make parsing back into the year/month/day for citation/CSL or even (in my case) manual generation, difficult. Doing this in twig is possible (basically anything is possible) but also too much engineering

What I propose is to create a new extension 

`edtf_2_iso_date` that uses a similar logic as the one found when sending ETDF to Solr. From a given EDFT string we process the lower and upper limits back to valid ISO8601 (without minutes, seconds, etc, formatted to "day" precision and deduplicated). If someone then needs to make those humanized the existing extension can be called against those returned values. That is more flexible

What do you think?

DiegoPino:
Solved. New twig extension (needs documentation). `edtf_2_iso_date` returns and array of valid ISO dates given a valid EDTF one. If something fails/the passed date is wrong, the return is an empty array
https://github.com/esmero/format_strawberryfield/commit/b028600d003c0ba0294020dfed436fc24fc82810

PR #370, opened 2023-09-29
ISSUE-368: edtf_2_iso_date

DiegoPino:
# What? see #368 

`edtf_2_iso_date` which is a filter returns always an array. Either empty if we could not parse it, wrong data, etc. a single 'Y-d-m' value or two  'Y-d-m' if a range/uncertainty with 2 possible dates.


DiegoPino:
Note: this pull contains also #366. Just bc whoever gets merged first wins (and if that gets merged first this one will be 1:1/no need to rebase)

Issue #385, opened 2023-12-06
Drupal 10.1 larger changes (JS/AJAX/Views GET instead of POST)


DiegoPino:
# What?

Drupal 10.0 is quite different to Drupal 10.1. Largest change is a rewrite of how Symfony 5 v/s Ajax (now uses GET in Views) works/fetches arguments from a controller. We do this a lot inside archipelago to drive our custom Exposed Form Blocks, our Facet Ajax implementations, etc. So rewriting is needed for many components in our submodules. This is for a new release. So 1.4.0 for sure

Also, I want to be sure I restrict (remove the combat) for Facets 3.0. Broken module so far, basically changes everything (facets become filters) without providing all the functionality we expect from that module

DiegoPino:
Solved via https://github.com/esmero/format_strawberryfield/commit/f53d1e16746cbe16bdb5dc2ffcd65f10bc84876c
and 
https://github.com/esmero/format_strawberryfield/commit/5072ecfd94128562064e5974b11936f7f5fb4a9a

PR #391, opened 2023-12-10
ISSUE-385: Drupal 10.1 compatibility for our overriden Controllers/facet/exposed forms (Symphony 5 v/s GET)

DiegoPino:
See #385 

This enables us to move forward with Drupal 10.1. Views now use by default GET which also means the arguments we need for our overrides for facets/ custom exposed form block approach are somewhere else. 

I'm not sure of these changes work with 10.0.x so for now will pull against 1.3.0 but might have to change my mind and roll YET another release. (thanks blue drop)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
         }
     ],
     "require": {
+	"drupal/core":"^10.1", 
         "strawberryfield/strawberryfield":"1.3.0.x-dev",
         "league/html-to-markdown":"^5.0.0",
         "erusev/parsedown": "^1.7",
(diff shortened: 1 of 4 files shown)


DiegoPino:
@alliomeria @WWU-MaxBronsema I tested this pull against 1.3.0 on Drupal 10.0.x. Does not work. The breaking change in Drupal 10.1 (Views using GET arguments) is backwards compatible with this. Means I will have to roll a release later this weekend. Will move fast and let you all know. It will be a 1.3.1 not a 1.4.0 (so we don't have to write new documentation). Thanks

PR #393, opened 2023-12-15
ISSUE-207-1.3 --- Moved twig context alter hook inside the render con…

patdunlavey:
See #392 for 1.2.0 PR
- Moved twig context alter hook inside the render context to avoid potential cache 'leakage' errors.

### DIFF
--- a/src/Controller/MetadataExposeDisplayController.php
+++ b/src/Controller/MetadataExposeDisplayController.php
@@ -270,20 +270,19 @@ class MetadataExposeDisplayController extends ControllerBase {
               'format_strawberryfield.iiif_settings'
             )->get('pub_server_url');
             $original_context = $context + $context_embargo;
-            // Allow other modules to provide extra Context!
-            // Call modules that implement the hook, and let them add items.
-            \Drupal::moduleHandler()->alter(
-              'format_strawberryfield_twigcontext', $context
-            );
-            // In case someone decided to wipe the original context?
-            // We bring it back!
-            $context = $context + $original_context;
-            $cacheabledata = [];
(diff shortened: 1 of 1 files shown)


src/Controller/MetadataExposeDisplayController.php
```diff
@@ -270,20 +270,19 @@ public function castViaTwig(
...
             $original_context = $context + $context_embargo;
-            // Allow other modules to provide extra Context!
-            // Call modules that implement the hook, and let them add items.
-            \Drupal::moduleHandler()->alter(
-              'format_strawberryfield_twigcontext', $context
-            );
-            // In case someone decided to wipe the original context?
-            // We bring it back!
-            $context = $context + $original_context;
-            $cacheabledata = [];
```
DiegoPino, line 281: @patdunlavey can you keep please  $cacheabledata = []; in line 281?

DiegoPino (changes requested):
Small change. Not a biggie but just bc in the future executeInRenderContext might change signature and not return an empty array. Thanks

src/Controller/MetadataExposeDisplayController.php
patdunlavey: I'm afraid I don't understand. Restoring it to a position before `$this->renderer->executeInRenderContext` will simply result in it being overwritten, and putting it after will clobber the result of `$this->renderer->executeInRenderContext`, and putting it inside the inline function does nothing that I can see.

patdunlavey: Nevermind. I've gone ahead and updated the issue branch with the change as requested (and as I understand it).

DiegoPino (approved):

Issue #395, opened 2023-12-21
Internal Page Cache does not "vary" for Anonymous users in 9.x, 10.0 and 10.1. IP embargoes needed that. Kill cache


DiegoPino:
# What?

Basically anonymous users have no Cache Context. They do not vary. So IP embargoes only work for Logged in users in a reliable way right now. Solution?
1.- Check if embargoes are enabled at all
2.- Check if user is anonymous
3.- If anonymous, check if IP data (for allowing/not allowing) is at all present in the ADO
4.- If present, check if the Internal Page Cache module is enabled (if not, the smart cache, which is aware but also slower, can deal with it, treating anonymous as a varying user) and Kill the Page Cache.
5.- Set any output after to max-age = 0 to ensure there is not even client level caching

This needs to be applied at a single point (good!) because we have a cache resolving service (good Diego)

@alliomeria for tomorrow

DiegoPino:
Solved via https://github.com/esmero/format_strawberryfield/commit/7e95fa6e5aff5166e99ab4b9abb17d9b79a7b7fa

DiegoPino:
I will also make a backwards compatible one for 1.2.0 soon

PR #396, opened 2023-12-22
ISSUE-395: Make IP based embargoed ADOs + Anonymous users un-cacheable

DiegoPino:
See #395 

This has been tested a lot. But I will give my today's self a bit of slack, test again later tonight and merge. Solves the issue with Anonymous users from a different IP/Network seen a stale cache generated by other anonymous users + makes also cache more robust on all viewers +  brings to 2023 the simpler image formatter nobody uses (a "classic")


DiegoPino:
This needs also a re-roll for 1.2.0

Issue #400, opened 2024-01-10
Advanced Search: "classic" mode.


DiegoPino:
# What?

We have been asked by the community to make advanced search to work closer to what older Library catalog/search systems used to work, means to automatic update on results when you add/remove terms. To do so I will code some extra options (config options) so I can intercept the adding/removing of a term but without triggering a search. This is really not easy at all since Drupal needs to process input to be able to return exposed filters and the number of them, so probably this will have to be handled as a JS task. This implies also that only pressing the "Search" button itself will trigger a search.

At the same time there is a request to have `minus` (or remove) buttons for each row, instead a global one at the end.

Additionally we have been requested to allow Facet Summaries for Full text to have separate "remove" buttons per field. Also a complex task, since the fact that our exposed filter has multiple entries (text) is not visible really to Drupal. Views Filters tend to be atomic. To do so, I will have to keep track of each Sub element's id and make sure people can't trick the View into accepting un-filtered user arguments.  This will be also an option, since by design we still find very useful the ability of clearing all "free text" search. This also means: removing a filter entry (means e.g the second row) will need to submit and refresh the view. We can't (or should we) act on "classic" mode for this, since Facet Summaries are their own module and we really have no control on the submit action on those.

@alliomeria please correct me if I'm reading this wrong.

DiegoPino:
Solved via https://github.com/esmero/format_strawberryfield/commit/657c28a66bdfab2e4a436d3e6421c916ab119351

PR #403, opened 2024-01-10
ISSUE-400: Advanced Search Classic mode

DiegoPino:
See #400 

This is ready to be tested.

New settings:

image

Classic mode basically uses JS for everything but keeps the sequential logic of fields named with a suffix `_$count` and ordered. Does not solve the other need (facet summaries)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,6 @@
         }
     ],
     "require": {
-	"drupal/core":"^10.1", 
         "strawberryfield/strawberryfield":"1.3.0.x-dev",
         "league/html-to-markdown":"^5.0.0",
         "erusev/parsedown": "^1.7",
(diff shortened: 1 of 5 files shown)


DiegoPino:
Notes for myself:

- On the Exposed filter I need to keep track if previously "search" was pressed and kind build a "bookmark" for that. Because on add/remove I still need to submit the form to the views. And if not, every action will trigger a search, but if I don't want that, and e.g I want to keep the previous result of a search visible (bc if not adding fields will blank the results right) I need to know what were the last submitted values. This can be a hidden field? Could be

alliomeria:
This is great @DiegoPino! So far all testing in local has been going very well, no issues using either modern/regular modality for the interactions or with the 'classic' mode setup.

Next week, will work with you to pass some potential default configurations (views, facets, etc) to pair with these for standard deployments.

Thank you and have a great weekend! :D

DiegoPino:
@alliomeria thanks so much for all the extensive testing. Same for you, happy weekend 🐶 🌷

Issue #404, opened 2024-01-22
Add option for individual entries (in specific for Advanced Search) for Full Text input facet Summaries


DiegoPino:
# What?

Related to #403 but not the same (bc a different submodule). Seems like there is a need to allow users to "unselect" a specific Full Text search entry from a facet summary instead of "clear" all Search Full text, which is right now the beaver (not a bug, this is how we designed it). This has many issues which I can not fix:
- The user (or us) can not know which Facet Entry on an Advanced Search is e.g the one that lead to no results. Which means deleting one might NOT give the user more results/achieve what is desired
- We can not allow gaps between fields. So if adv_1 and adv_2 and adv_3 are passed and the user deletes adv_2, we need to move adv_3 to adv_3, means we need to, on every facet summary read how the Filter itself is configured to iterate from 1 to N allowed fields and re-build them. This is hard.

But it was requested, so will try it out.

DiegoPino:
Solved via https://github.com/esmero/format_strawberryfield/commit/1ab346f01aaeb7b2713a45ff8fc59baae7a2325e

PR #405, opened 2024-01-22
ISSUE-404: Facet summary with separate entries per Input/field for advanced search

DiegoPino:
See #404 . I will keep pulling this against 1.4.0 but maybe we can just make all 1.3.0 and then close the Restaurant, turn the tables and wash the dishes?

Who knows

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,6 @@
         }
     ],
     "require": {
-	"drupal/core":"^10.1", 
         "strawberryfield/strawberryfield":"1.3.0.x-dev",
         "league/html-to-markdown":"^5.0.0",
         "erusev/parsedown": "^1.7",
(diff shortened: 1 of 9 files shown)


DiegoPino:
@alliomeria will do some testing now. Not sure yet if my logic works/makes sense. LOTs of array/moving things around, etc. If this works on my side I will ask you to test also during the week? Before the end of the week (bc I need to update a server during the weekend with this). Appreciate all your energies and efforts on this

DiegoPino:
@alliomeria one issue I noticed - tomorrow but have to type now or I will forget- (which is not our processor I think?) is that if the facet block is set to OR (bad idea for advanced search) and we have 0 results, the current passed facet (e.g my own name) is not available to our processor (bc it is not part of the list) but also bc Drupal takes over the Facet summary (given than the facet block itself shows all the others to allow for an OR) it skips my logic and thus not showing the facet summary for the one that fails (my name, I fail)... hard to explain, will have to show tomorrow. Will double check the code if I can intercept/override this from the main Facet Summary. Seems like a bug to me

Issue #408, opened 2024-01-31
Advanced search and "required" filter combination requires a different approach 


DiegoPino:
# What?

To allow advanced search to start with an empty listing we need to make the filter "required". When we do so, the first Search (by inheritance) is used to check for a value. If it is empty, the whole query is cancelled. This does not take in account the option that maybe other fields are filled in with. To add even more complexity, we need to also do extra check for character length, but skip empty fields if at least one is not empty.

Just more complex stuff to make people happy.

DiegoPino:
Resolved all over the place via https://github.com/esmero/format_strawberryfield/commit/b2bacba423d1e6643209209d4c4da43113b7fb3f

PR #409, opened 2024-01-31
ISSUE-408: Adds multi value check for required, validate count on non empties

DiegoPino:
See #408


DiegoPino:
@alliomeria for completeness I will also merge another pull of this against 1.3.0 so you can test what I showed you (an empty first field but data in the other ones). But first I will make sure 1.4.0 works. Thanks

PR #415, opened 2024-02-13
Issue-414--leaflet-options --> 1.3.0

patdunlavey:
### What
- Added leaflet map option for 'Disable mouse wheel zoom'. 
- Reduced the minimum allowable leaflet map height to 62px, which guarantees that the +/- buttons will display while permitting map heights smaller than 480 pixels.

### To Test:
- Edit "Map Display for Digital Objects SQL view (`/admin/structure/views/view/map_display_for_digital_objects_sql`) view
- Click on the field settings for "Content: 🍓 Strawberry (Descriptive Metadata source)" to bring up the configuration form for that field.
- Test the view with different values for "Maximum height". Any height equal to or larger than 62 pixels should be able to be saved, and should be reflected in height of the map that is displayed when the view is rendered for an object.
- Set the "Disable mouse wheel zoom" checkbox to unchecked (it's the default) and save the view. When viewing the map, scrolling the mouse wheel should cause the map to zoom in and out.
- Check the "Disable mouse wheel zoom" checkbox and save the view. Now, using the mouse wheel to scroll the page that the map is rendered on should not cause the map to zoom when the cursor enters the map area.

### DIFF
--- a/js/leaflet_strawberry.js
+++ b/js/leaflet_strawberry.js
@@ -46,8 +46,12 @@
             if (drupalSettings.format_strawberryfield.leaflet[element_id]['initialzoom'] || drupalSettings.format_strawberryfield.leaflet[element_id]['initialzoom'] === 0) {
               $initialzoom = drupalSettings.format_strawberryfield.leaflet[element_id]['initialzoom'];
             }
+            var $scrollWheelZoom = !drupalSettings.format_strawberryfield.leaflet[element_id]['disable_mouse_zoom'];
             // initialize the map
-            var map = L.map(element_id).setView([40.1, -100], $initialzoom);
+            var map = L.map(
+              element_id,
+              {scrollWheelZoom: $scrollWheelZoom}
+            ).setView([40.1, -100], $initialzoom);
             // Use current's user lat/long
             // Does not work without HTTPS
             //  map.locate({setView: true, maxZoom: 8});
--- a/js/leaflet_strawberry.js
+++ b/js/leaflet_strawberry.js
@@ -46,7 +46,14 @@
             if (drupalSettings.format_strawberryfield.leaflet[element_id]['initialzoom'] || drupalSettings.format_strawberryfield.leaflet[element_id]['initialzoom'] === 0) {
               $initialzoom = drupalSettings.format_strawberryfield.leaflet[element_id]['initialzoom'];
             }
-            var $scrollWheelZoom = !drupalSettings.format_strawberryfield.leaflet[element_id]['disable_mouse_zoom'];
+
+            var $scrollWheelZoom = true;
+
+            if( typeof drupalSettings.format_strawberryfield.leaflet[element_id]['disable_mouse_zoom'] !== 'undefined') {
+              // Use the "truthy" value of the disable_mouse_zoom setting to determine if we should disable scroll wheel zoom
+                $scrollWheelZoom = !(!!drupalSettings.format_strawberryfield.leaflet[element_id]['disable_mouse_zoom']);
+            }
+
             // initialize the map
             var map = L.map(
               element_id,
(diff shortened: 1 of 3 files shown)


patdunlavey:
@DiegoPino shall I make a 1.4.0 PR too?

DiegoPino:
Yes please. 1.3.0 should be closed already (but i’m not such a strict human
so there is that)

js/leaflet_strawberry.js
```diff
@@ -46,8 +46,12 @@
             if (drupalSettings.format_strawberryfield.leaflet[element_id]['initialzoom'] || drupalSettings.format_strawberryfield.leaflet[element_id]['initialzoom'] === 0) {
               $initialzoom = drupalSettings.format_strawberryfield.leaflet[element_id]['initialzoom'];
             }
+            var $scrollWheelZoom = !drupalSettings.format_strawberryfield.leaflet[element_id]['disable_mouse_zoom'];
```
DiegoPino, line 49: @patdunlavey this needs a fall back default. Since it is a new setting and some users might not have updated their configs (and there is no update hook, pretty sure it is not needed also) you should check if the value is defined at all first and test

src/Plugin/Field/FieldFormatter/StrawberryMapFormatter.php
```diff
@@ -476,10 +477,11 @@ public function settingsForm(array $form, FormStateInterface $form_state) {
           '#type' => 'number',
           '#title' => $this->t('Maximum height'),
           '#default_value' => $this->getSetting('max_height'),
+          '#description' => $this->t('The minimum permitted value is 62 pixels to ensure that the zoom controls are visible on the map.'),
           '#size' => 5,
           '#maxlength' => 5,
           '#field_suffix' => $this->t('pixels'),
-          '#min' => 0,
+          '#min' => 62,
```
DiegoPino, line 484: Ok.. so we already had min 0 here ... so the original use case was solved? Won't fight it

DiegoPino (changes requested):
Schema YAML file still needs to be modified for this formatter + the JS needs to check for undefined (please test given that this is a nested setting in JS you can't just access a key if it does not exist) and default to FALSE if not present. Thanks!

patdunlavey:
Thanks Diego! I'll get this revised today.

js/leaflet_strawberry.js
patdunlavey: I rewrote this to use something like the methodology used with other settings. It now...
- sets a default for `$scrollWheelZoom` (true)
- checks if `drupalSettings.format_strawberryfield.leaflet[element_id]['disable_mouse_zoom']` is defined
- if so, sets `$scrollWheelZoom` to the "truthy" version of that value (negated)

src/Plugin/Field/FieldFormatter/StrawberryMapFormatter.php
patdunlavey: Just making it match what we enforce later.

js/leaflet_strawberry.js
patdunlavey: I added
```
    disable_mouse_zoom:
      type: string
      label: "Disable mouse wheel zoom"
```
Let me know if that's not correct (not sure about `type: string`).

DiegoPino:
@patdunlavey note:
I tested with boolean and the output after reseting the settings, saving, re-setting them is this
```YML
            min_zoom: 2
            disable_mouse_zoom: true
            tilemap_url: 'https://a.tile.openstreetmap.org/{z}/{x}/{y}.png'
```

and all works well.

patdunlavey:
format_strawberryfield.schema.yml change pushed on both issue branches: changed "string" to "boolean".

Though I did not entirely remove and recreate the view config, I did `drush cdel views.view.map_display_for_digital_objects_sql display.default.display_options.fields.field_descriptive_metadata.settings.disable_mouse_zoom`, which I believe re-creates the config state prior to the code change. I tested that it used the default (mouse wheel zoom enabled) without any errors or warnings. 

Thanks for your review of this @DiegoPino !

DiegoPino (approved):
@patdunlavey thanks so much. Tested and all works well

patdunlavey:
Thank you!

PR #416, opened 2024-02-13
Issue-414-1.4.0--leaflet-options --> 1.4.0

patdunlavey:
### What 
1.4.0 version of [this PR](https://github.com/esmero/format_strawberryfield/pull/415)
- Added leaflet map option for 'Disable …mouse wheel zoom'. 
- Reduced the minimum map height to 62px, which guarantees that the +/- buttons will display.

### To Test:
- Edit "Map Display for Digital Objects SQL view (`/admin/structure/views/view/map_display_for_digital_objects_sql`) view
- Click on the field settings for "Content: 🍓 Strawberry (Descriptive Metadata source)" to bring up the configuration form for that field.
- Test the view with different values for "Maximum height". Any height equal to or larger than 62 pixels should be able to be saved, and should be reflected in height of the map that is displayed when the view is rendered for an object.
- Set the "Disable mouse wheel zoom" checkbox to unchecked (it's the default) and save the view. When viewing the map, scrolling the mouse wheel should cause the map to zoom in and out.
- Check the "Disable mouse wheel zoom" checkbox and save the view. Now, using the mouse wheel to scroll the page that the map is rendered on should not cause the map to zoom when the cursor enters the map area.

### DIFF
--- a/config/schema/format_strawberryfield.schema.yml
+++ b/config/schema/format_strawberryfield.schema.yml
@@ -662,6 +662,9 @@ field.formatter.settings.strawberry_map_formatter:
     min_zoom:
       type: integer
       label: 'Min Zoom for the Map'
+    disable_mouse_zoom:
+      type: string
+      label: "Disable mouse wheel zoom"
     tilemap_url:
       type: string
       label: 'Tile Map URL'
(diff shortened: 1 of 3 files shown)


patdunlavey:
Updated after review on the [1.3.0 companion PR](https://github.com/esmero/format_strawberryfield/pull/415).

DiegoPino (approved):
Great! Thanks @patdunlavey

patdunlavey:
Woohoo!

Issue #292, opened 2023-04-06
IAbookreader: Make paging mode/direction/cover all be driven by IIIF


DiegoPino:
#What?

We are already using IIIF to drive this. Allow manifest level/canvas level paging hints to drive the more limited options (but still present) on IAbookreader.

This is basically JS, parsing the properties we need from the IIIF V2 and V3 presentation API into the matching code in IAbookreader

I wrote this already a time ago. So will reuse my code (very own, not copying anyone's code here) + add some extras coming from IIIF V3 now

DiegoPino:
@patdunlavey this is the code I wrote that we need to port into a IIIF driven paging hint functionality 

https://github.com/mnylc/islandora_internet_archive_bookreader/commit/ab054fa63b87bf7802a429582566ab7368a8e071#diff-25b528e228dae15a98036e6015863597bb5ae2a24c1705c960e0d36ed07396b4R21-R30

patdunlavey:
Great, thanks @DiegoPino . 

I'm not sure how you are thinking that I7 idea sheds light on this. In that code there is a function `islandora_paged_content_get_viewing_hint` to determine whether we want 1up or 2up. It first looks to see if the object has a viewing hint in RELS_EXT, and if not, then checks for a config set by the islandora_bookreader module. In other words, it's all about how you compute what display mode IAB should use. It seems to me that the idea of putting this into the IIIF Manifest metadatadisplay template essentially offloads that question to the site builder - it's however the template writer chooses to decide it. Am I right?

What I don't understand is how you are thinking the object and canvas level view mode hinting should be encoded in the manifest (see next), and once that's done, how `BookReader.loadManifest` will use that information to instruct BookReader.

In the IIIF manifest, are you thinking whatever logic is being used would be rendered out into the manifest as something like this?
```
....
"sequences": [
  {
    ....
    "viewingHint": "paged",
    "bookreaderViewingHint": "1up" {# "1up", "2up", or "thumb"#}
    ....
  }
],
...
"canvases": [
  {
    ....
    "bookreaderViewingHint": "2up" {# "1up", "2up", or "thumb"#}
    ....
  }
],
....
```

DiegoPino:
Hi, thanks for following up. I will share the IIIF manifest canvas official usage of this hints for v2 and v3 before the end of the day. The heavier part for us here will be the JS part to extract the info

More soon!

Diego Pino Navarro
Director of Digital Strategy
Archipelago architect
[...]

PR #419, opened 2024-02-21
ISSUE-292: IIIF driven paging/sequence + order drive hints + cover page

DiegoPino:
See #292 

@alliomeria and @pcambra don't test yet, I need to add a few extra config options for the formatter too. Thanks

### DIFF
--- a/js/plugin.iiif-iabookreader_strawberry.js
+++ b/js/plugin.iiif-iabookreader_strawberry.js
@@ -200,6 +200,7 @@ BookReader.prototype.setup = (function(super_) {
     this.searchInsideUrl = options.searchInsideUrl;
     this.enableSearch = options.enableSearch;
     this.goToFirstResult = false;
+    this.hasCover = false;
 
     // Base server used by some api calls
     this.bookId = options.bookId;
@@ -219,6 +220,47 @@ BookReader.prototype.init = (function(super_) {
     super_.call(this, options);
   };
 })(BookReader.prototype.init);
+
+/**
+ * Return which side, left or right, that a given page should be
+ * displayed on. override allows a no-cover situation.
+ */
+BookReader.prototype.getPageSide = function(index) {
+  if ('rl' != this.pageProgression) {
+    // If pageProgression is not set RTL we assume it is LTR
+    if (0 == (index & 0x1)) {
+      // Even-numbered page
+      if (false === this.hasCover) {
+        return 'L';
+      }
+      return 'R';
+    }
+    else {
+      // Odd-numbered page
+      if (false === this.hasCover) {
+        return 'R';
+      }
+      return 'L';
+    }
+  }
+  else {
+    // RTL
+    if (0 == (index & 0x1)) {
+      if (false === this.hasCover) {
+        return 'R';
+      }
+      return 'L';
+    }
+    else {
+      if (false === this.hasCover) {
+        return 'L';
+      }
+      return 'R';
+    }
+  }
+}
+
+
 
 BookReader.prototype.getApiVersion = function() {
   var self = this;
@@ -252,7 +294,7 @@ BookReader.prototype.loadManifest = function () {
     self.bookUrl = '#';
     // self.thumbnail = self.jsonLd.thumbnail['@id'];
     self.metadata = self.jsonLd.metadata;
-    // Assuming i have no default sequence so first one is the good one.
+    // Assuming we have no default sequence so first one is the good one.
     self.parseSequence(null);
   }
   else {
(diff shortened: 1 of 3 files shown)


DiegoPino:
@alliomeria and @patdunlavey this is ready for review:

What is new (or ho to test). Get the code of course (this is based on 1.4.0)
- Modify your IIIF Manifests (works with V3 and V2) to use the following properties at the top level of a IIIF Manifest (means this does not work with structures or sequence, remember this is a book reader, not a fully IIIF aware viewer like mirador

- for iiif v2, at the top level of the manifest, either:
"viewingHint": "paged"
or
"viewingHint": "individuals"
for v3
also at the top level of the manifest
"behavior":["paged"], // Notice the array here OR
"behavior":["individuals"],
mirador will also respect that
For viewing direct V2 and V3)
"viewingDirection" : "right-to-left"
OR
"viewingDirection" : "left-to-right"

for the cover (see the explanation on the new setting for this formatter, set a RAW JSON key with either true/false values in the ADO itself
will default to "hasCover": true (normal book heavier)

This pull also brings exposed metadata displays to the IABookreader. This is better than the internal Metadata display because it will allow IABookreader to adapt without a cache reset (because the actual URL is the one dealing with the cache)

I might add more cache goodies once you test this ok?

Thanks you both

alliomeria:
Awesome, thank you @DiegoPino !

Following up to note the issue we talked about internally earlier today related to the placement of the viewingHint for v2. Technically, valid IIIF v2 specs permit placement of viewingHint nested within "sequence", but for this iteration of the PR, it needs to be outside/out a level.

Will check on the cover image setup also and let you know if I have any questions or feedback about that.

Thanks again, Diego! Think people are going to groove on the extra options you're providing. :)

DiegoPino:
@alliomeria thanks for the feedback. Will make sure I also take the sequence in consideration, I think our default shipped V2 one has it there, so it is a must. :)

DiegoPino:
@alliomeria code/issue addressed! Thanks. @patdunlavey can you please also give this a try?spin? Would love to merge as soon as possible if all works given that there is another open pull against the same code that will need a rebase from me. Thanks a lot

patdunlavey:
@DiegoPino thank you for all your work on this!! I will try to test it out tomorrow.

DiegoPino:
@patdunlavey I'm rebasing this pull (because of #420 ) so if testing don't try to merge against 1.4.0. Just fetch this branch and test.  Thanks

DiegoPino:
@patdunlavey and @roromedia. This was rebased and includes also the code from #420. test at will. If I don't get from anyone "this does not work" I will merge tomorrow by noon. Thanks!

patdunlavey:
@DiegoPino thanks,  I tested this with both API 2.1 and 3.0 manifests. After some futzing around with the metadatadisplay templates and the display settings, I got the `"viewingHint": "individuals"` (v2.1 manifest) working, and I got `"behavior": ["individuals"]`(v3.0 manifest) working. That was a while ago. However now it's not working and I haven't figured out why. 

After the initial success with the individuals view, I set about trying to figure the "hasCover" option. I set  ` "hasCover": 0,` in the json for an object and configured both manifest versions to specify the "paged" mode. I assumed that this should result in the first page being displayed 2-up (page 1 and 2), but I did not see that. It just displayed a single page in the right half of the viewer for page 1.

Subsequently I tried to re-create the 1-up ("individuals") configuration that I had working earlier, and so far no amount of futzing with the v2.1 and v3.0 templates or display settings has given me back my vertical scrolling single page view. I've performed regular drupal cache-clear repeatedly, as well as switching to incognito and a different browser, and the single page view refuses to return. Presumably I missed something in my trips through the configs, but I'll be damned if I can see what it is.

DiegoPino:
Ok, let me double check (I just checked but maybe my rebase just messed up stuff). Did you clear also your browser caches? Any errors on your browser JS console.

DiegoPino:
Will share on slack a tiny video ok @patdunlavey ?

patdunlavey:
:man_facepalming: I found the problem - "individual" vs "individuals"! NEVERMIND!

DiegoPino:
Good! Ok, will still make the video bc I love movies! Will share soon! hahahah

patdunlavey:
And I figured out my "hascover" problem. Once again an itty-bitty typo! "hasCover" is not the same as "hascover"! :man_facepalming: :fist_left: 

With these two things working, I think that complete my review. I'll indicate such with a proper review comment.

config/schema/format_strawberryfield.schema.yml
```diff
@@ -308,6 +308,11 @@ field.formatter.settings.strawberry_paged_formatter:
       type: string
     manifesturl_source:
       type: string
+    metadataexposeentity_source:
```
patdunlavey, line 311: Once I figured out that upper/lower case matters in the `hascover` field name, I confirmed that setting it to false, using the 'paged' view hint, I get two pages displaying side by side in the default view. :+1:

DiegoPino:
Thanks a lot @patdunlavey key sensitive, yep a thing in JSON (funny not in PHP ... but yes if an array key... anyways.... appreciate the testing! Video shared too. hugs

js/plugin.iiif-iabookreader_strawberry.js
```diff
@@ -288,27 +330,80 @@ BookReader.prototype.loadManifest = function () {
 BookReader.prototype.setupTooltips = function() {
 };
 
+function processBehaviorToIABookreaderModeV3(jsonLd) {
```
patdunlavey, line 333: Tested with
```
"behavior":["paged"],
```
and
```
"behavior":["individuals"],
```
In a IIIF v3.0 manifest, and I can confirm this is working.

js/plugin.iiif-iabookreader_strawberry.js
```diff
@@ -288,27 +330,80 @@ BookReader.prototype.loadManifest = function () {
...
+      if (jsonLd.behavior.includes("individuals")) {
+        return 1;
+      }
+    }
+  }
+
+return 2;
+}
+
+function processBehaviorToIABookreaderModeV2(jsonLd) {
```
patdunlavey, line 360: Tested with 
```
"viewingHint": "paged",
```
and 
```
"viewingHint": "individuals",
```
in a IIIF v2.1 manifest metadatadisplay template and I can confirm this is working.

js/plugin.iiif-iabookreader_strawberry.js
```diff
@@ -288,27 +330,80 @@ BookReader.prototype.loadManifest = function () {
...
+  // Bookreader supports only two modes but also a book cover or not
+  // IIIF Presentation API V2 and V3 supports more
+  // So we need to reduce/map back to only 2 options
+  // V3 key at manifest level/and/or Sequence level is `behavior`
+  // V2 key at manifest level/and/or Sequence is `viewingHint`
+  // normally we would use https://github.com/internetarchive/bookreader/blob/4dab4cef30c0af05fa864e57578a834b89fbaba2/src/BookReader.js#L62
+  // but these have not changed in 10 years ... passing this would be an overkill just to read that.
+  if (typeof(jsonLd.viewingHint) !== "undefined") {
+    console.log (jsonLd.viewingHint);
```
patdunlavey, line 369: Remove this console.log statement? Only needed for debugging, I assume.

patdunlavey (changes requested):
Only one change request - to remove a debugging statement in `plugin.iiif-iabookreader_strawberry.js`

js/plugin.iiif-iabookreader_strawberry.js
DiegoPino: Removed. Thanks for catching that!

patdunlavey (approved):
Thanks for your wonderful work on this Diego!!

DiegoPino:
Thanks to you @patdunlavey !

PR #420, opened 2024-02-22
Adds the ability to change the default base URL for the images of IA BookReader

roromedia:
This is necessary as because of GDPR as it is not allowed to load data from external sources without consent of the user. Now by making the path configurable the user is able to self-host these images.

@DiegoPino : There is for sure a more standardised way to put the default configuration path. I now added a function which returns this path but I guess it would be better retrieved out of a settings-file or via \Drupal::config()?

### DIFF
--- a/js/iiif-iabookreader_strawberry.js
+++ b/js/iiif-iabookreader_strawberry.js
@@ -40,7 +40,7 @@
                         iiifmanifest: strawberrySettings['manifest'],
                         iiifdefaultsequence: null, //If null given will use the first sequence found.
                         maxWidth: 800,
-                        imagesBaseURL: 'https://cdn.jsdelivr.net/gh/internetarchive/bookreader@4.40.3/BookReader/images/',
+                        imagesBaseURL: strawberrySettings['iareaderimagesbaseurl'],
                         server: server,
                         bookId: node_uuid,
                         enableSearch: true,
--- a/js/iiif-iabookreader_strawberry.js
+++ b/js/iiif-iabookreader_strawberry.js
@@ -1,72 +1,72 @@
 (function ($, Drupal, once, drupalSettings) {
 
-    'use strict';
+  'use strict';
 
-    Drupal.behaviors.format_strawberryfield_iabookreader_initiate = {
-        attach: function(context, settings) {
+  Drupal.behaviors.format_strawberryfield_iabookreader_initiate = {
+    attach: function(context, settings) {
 
-          const elementsToAttach = once('attach_iab', '.strawberry-iabook-item[data-iiif-infojson]', context);
-          $(elementsToAttach).each(function (index, value) {
-              // Get the node uuid for this element
-              var element_id = $(this).attr("id");
-              var strawberrySettings = drupalSettings.format_strawberryfield.iabookreader[element_id];
-              var server = window.location.origin + '/';
-              let textselection = false;
+      const elementsToAttach = once('attach_iab', '.strawberry-iabook-item[data-iiif-infojson]', context);
+      $(elementsToAttach).each(function (index, value) {
+        // Get the node uuid for this element
+        var element_id = $(this).attr("id");
+        var strawberrySettings = drupalSettings.format_strawberryfield.iabookreader[element_id];
+        var server = window.location.origin + '/';
+        let textselection = false;
 
-              // Check if we got some data passed via Drupal settings.
-              if (typeof(strawberrySettings) != 'undefined') {
-                var node_uuid = strawberrySettings['nodeuuid'];
-                if (typeof(strawberrySettings['server']) != 'undefined') {
-                  server = strawberrySettings['server'];
+        // Check if we got some data passed via Drupal settings.
+        if (typeof(strawberrySettings) != 'undefined') {
+          var node_uuid = strawberrySettings['nodeuuid'];
+          if (typeof(strawberrySettings['server']) != 'undefined') {
+            server = strawberrySettings['server'];
+          }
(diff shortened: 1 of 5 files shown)


DiegoPino:
@roromedia thanks! could you please remove the two dot files from this pull? I will test in the meantime and give you feedback if some changes are needed (If at all). Thanks!!!

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -127,6 +127,7 @@ public static function defaultSettings() {
       'manifesturl_source' => 'iiifmanifest',
       'max_width' => 720,
       'max_height' => 480,
+      'ia_reader_images_base_url' => '',
```
DiegoPino, line 130: Formatters using the ::getSetting method will eventually call this one here

see
```public function getSetting($key) {
    // Merge defaults if we have no value for the key.
    if (!$this->defaultSettingsMerged && !array_key_exists($key, $this->settings)) {
      $this->mergeDefaults();
    }
    return $this->settings[$key] ?? NULL;
  }
```
Which means by setting the return of getIABookReaderImagesDefaultBaseURL() directly you can save yourself that function and you can also call the defaultSettings method directly https://github.com/esmero/format_strawberryfield/pull/420/files#diff-b23e313798ac690e86b6697e23a4b490f091ea3b026b9480def3634656daf722R217 there instead of the private function

js/iiif-iabookreader_strawberry.js
```diff
@@ -40,7 +40,7 @@
                         iiifmanifest: strawberrySettings['manifest'],
                         iiifdefaultsequence: null, //If null given will use the first sequence found.
                         maxWidth: 800,
-                        imagesBaseURL: 'https://cdn.jsdelivr.net/gh/internetarchive/bookreader@4.40.3/BookReader/images/',
+                        imagesBaseURL: strawberrySettings['iareaderimagesbaseurl'],
```
DiegoPino, line 43: This JS here needs to check if strawberrySettings['iareaderimagesbaseurl'] is defined  (typeof !== "undefined")
before using it directly. You (sadly) will have to hardcode also (as a fallback) the https://cdn.jsdelivr.net/gh/internetarchive/bookreader@4.40.3/BookReader/images/' string here. Why? Because after updating the code many other Archipelago users will have still their Display Modes Cached, but the JS will update immediately which means that new variable will not be passed at all until someone either clears caches (expensive on a large repo) or save the formatter settings manually

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
```diff
@@ -407,6 +422,7 @@ public function processElementforMetadatadisplays($delta, array $jsondata, Field
     $max_width = $this->getSetting('max_width');
     $max_width_css = empty($max_width) || $max_width == 0 ? '100%' : $max_width .'px';
     $max_height = $this->getSetting('max_height');
+    $ia_reader_images_base_url = $this->getSetting('ia_reader_images_base_url');
```
DiegoPino, line 425: The logic on 'iareaderimagesbaseurl' => strlen($ia_reader_images_base_url) > 0 ? $ia_reader_images_base_url : $this->getIABookReaderImagesDefaultBaseURL(),
Should go here. And you can also check/add/strip/restore the last "/" to be sure it always ends in a single "/" in the same line.

DiegoPino (changes requested):
@roromedia I added a few comments. You can simply the code and add a bit of more validation/checks for existence in your pull still. Also, there is one entry missing from your pull which is the addition of a new config entry and a default after this line https://github.com/esmero/format_strawberryfield/blob/cf23b5879aef8d8621ed7cc18c89e78e72370cef/config/schema/format_strawberryfield.schema.yml#L320-L322

let me know if you have any questions. Almost there! Thanks!

src/Plugin/Field/FieldFormatter/StrawberryPagedFormatter.php
roromedia: I added the default url to the format_strawberryfield.schema.yml but I wasn't able to reuse or access it as the default value in the defaultSettings() function so I hardcoded the url there again.

js/iiif-iabookreader_strawberry.js
```diff
@@ -227,7 +227,7 @@ BookReader.prototype.buildToolbarElement = (function (super_) {
     var $el = super_.call(this);
     var readIcon = '';
     $el.find('.BRtoolbarRight').append("<span class='BRtoolbarSection Islandora'>"
-      + "<button class='BRpill zoomPage js-tooltip' title='Fine zoom'>ZOOM</button>"
+      + "<button class='BRpill zoomPage js-tooltip text-uppercase' title='Fine zoom'>Zoom</button>"
```
DiegoPino, line 230: Great idea!

DiegoPino:
@roromedia thanks for addressing the issues. Because of the spacing changes (which are good) I will have to actually git merge this against my local 1.4.0 to be able to see if there are any hidden issues. So please give me until tomorrow noon (EST) to have this properly tested. Thanks again

DiegoPino (approved):
@roromedia approving this. Since I have this larger pull open https://github.com/esmero/format_strawberryfield/pull/419 that touches many parts of the same code (and extensively) I will have to modify your code too before merging mine, so please expect a larger moving around of your contributed code during the next few days (will try to preserve as much as I can). Please make sure you do yet again a pull against 1.4.0 when that happens. Thanks

PR #426, opened 2024-03-15
ISSUE-425 --- Fixes brackets to get settings into JS

roromedia:


DiegoPino (approved):
Makes sense!

DiegoPino:
See #426 (@roromedia please always add a reference to the ISSUE so people can follow in the future what this fixes/where the explanation is. Thanks!)

Issue #427, opened 2024-03-15
Load Views via Ajax using data selectors provided by Archipelago


DiegoPino:
# What?

Many times one (you/me/them) need to load dynamically based on interaction or just async on load a list of things (a View). Might be because the list depends on conditions set on a Metadata Display with arguments, or on interaction (clicks). Writing code to do that is a bit advanced and might require rewiring some stuff. But all our users have access to editing their twig templates, so following the Bootstrap idea, let's use `data-` selector to give Ajax/JS commands on what can be done.

The idea is simple:

You output this:
```HTML
<button id="aRequiredID" 
data-sbf-view-id="creative_work_series_children" 
data-sbf-view-display-id="block_1" 
data-sbf-view-arguments="45" 
data-sbf-view-render-target="the_target" 
data-sbf-view-event="click">
Click me to load creative_work_series_children View 
with block_1 display 
passing argument 45 
inside another DIV with a class named "js-view-dom-id-the_target"
</button>

<div class="js-view-dom-id-the_target">HERE WILL BE MY VIEW LOADED</div>
```

DiegoPino:
resolved via https://github.com/esmero/format_strawberryfield/commit/8984758406ad600ee9846cc643a8d86a5d00a27e

PR #428, opened 2024-03-18
ISSUE-427: Adds data property (HTML) driven on click or on load Views loading via AJAX

DiegoPino:
see #427

This includes now also:

- Override for poorly coded thing in Drupal that does not allow nested Ajax Views Pagers
- Option to allow an on-click (`data-sbf-view-event="click"`)  or on-load (`data-sbf-view-event="load"`) for loading dynamically Views
- Works perfect. Just don't use the mini-pager because there is something wrong with it? Like globally (not us/but on them)

Tested with dozens of children, each child with 3K children. Paging works. All works!

### DIFF
--- a/modules/format_strawberryfield_views/format_strawberryfield_views.libraries.yml
+++ b/modules/format_strawberryfield_views/format_strawberryfield_views.libraries.yml
@@ -8,6 +8,7 @@ leaflet_strawberry_views:
     - core/drupal
     - core/drupalSettings
     - core/views
+    - views/views.ajax
     - format_strawberryfield/leaflet_ajax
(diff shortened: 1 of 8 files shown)


WWU-MaxBronsema:
This seems slick. Way to go!

alliomeria:
This is great @DiegoPino! So many Archipelago users will really benefit from this extra functionality. :) Thank you very much for your time coding this for everyone!

DiegoPino:
@alliomeria won't merge until I have at least basic docs. Might provide (the shades/colors and hues align) by Monday. Thanks

Issue #436, opened 2024-04-03
Replace PDF.JS 2.x with 4.x. && Learn how to use Node Modules (ES6) on web


DiegoPino:
# What?

I am old(ish) and using .mjs is not as intuitive in a Drupal environment as expected. .mjs are "modules" and are loaded differently. And even if Drupal allows us to do so (with a hidden library property/subproperty) there is nothing that ensure that a module will be available (the exports) by the time Drupal's behaviors kick in.

So ... async magic/reading 4 hours of messages of cool kids using nodejs and I found a way. Why also upgrading?

Found out that PDF.JS (2.x, one we had) could not read PDFs with "small defects" like many coming from the Internet Archive and older repos. Now those work. Pull coming tomorrow 

- Updates: the most simplistic nav we have is not accessible and can be accessed via tabs too
- Working on adding search: First PDF/JS native one using a different class. Then IIIF Content Search Driven (since this is rendering a canvas, I can render on top of it annotations)... and we can use the IIIF manifest a normal PDF would generate... more settings. This will lead to a "Search IIIF results component/element we can reuse in other viewers)
- Add a wrapper around the rendered canvas. That way we can render FULL but also make the viewer less tall/less width and add scroll bars (this one for @alliomeria bc it allows me to add a grey background, more pdf-profesh right?)

DiegoPino:
@patdunlavey I have no idea. Could you share the accessibility issues there? Are those things we/could tap in?

DiegoPino:
@patdunlavey i would love to have those as soon as possible, I need to move into other tasks (paging on IABookreader) + date Facet widgets and I fear things might fall out of radar. Will wait until tomorrow before making my pull.

mlohnash:
@DiegoPino and @patdunlavey 

I have made an excerpt from the recent CA-R accessibility audit with only the identified PDF.JS issues for Diego.

https://docs.google.com/spreadsheets/d/1A-gVddjGCrSRLkKzERFu9QlEk05MKIjf/edit?usp=sharing&ouid=101266773125396074198&rtpof=true&sd=true

DiegoPino:
@mlohnash thanks ❤️ . 

Ok this is important/useful. 

Here is the issue. Archipelago does not provide the full PDF.js interface you are using. Our interface is super simplistic. Next/previous page and full render. All the rest you have there, e.g 

https://californiarevealed.org/do/63881621-423b-4b03-918e-48b4934431c1

[...]

PR #437, opened 2024-04-04
ISSUE-436:  PDFJS 4.x + VTT Content Search API + Mirador 4.0 alpha 2 + ML

DiegoPino:
See #436 

- New ML supporting Code for Annotations + JS Triggers
- Newest PDF.js via module import/async (might need to pass the actual paths from PHP, but this also works)
- Theme level/template level accessibility and margins. Anyone can override the template
- New Option (simple/not so simple). Once I have the code the not so simple will enable native PDF.js search by using the pdfViewer class and the bus/dispatch/search controller. UPDATE: Or maybe I use the PDF Search Controller directly without the viewer. That would give me more control.
- Mirador 3.3.0 (default) and Mirador 4.0 alpha 2. (The former has CSS issues by the way, z-index)
- IIIF Content Search API with VTT and Plain Text (time/space mutation) processing + new settings
- JS improvements + small CSS
- New Twig template based Ajax/JS processors (data properties) for Views (on load/on click) and for Annotations (miniOCR) saved in Flavors with new Solr based Endpoint in our Annotations REST API


DiegoPino:
@alliomeria Merging. Will keep working on the new issue/features in other branches, this one just became too huge.
TODO:
- Document the new data-property (twig/html) driven Ajax/JS interactions. (for us too bc I will forget how this all works). I will make a pull request tomorrow AM + OpenAPI one too

Issue #442, opened 2024-04-22
More sanity checks on Open API Builder


DiegoPino:
# What?

See #394 also

On live/production test @alliomeria detected a few issues with Form Validation  (enumerations and UUID Entity validation) + Controller output (on empty we don't return a Response at all)

Will open a pull in a while to keep refining this new tool

alliomeria:
Thank you @DiegoPino, for safely catching 🐛 and releasing 🦋. 

I will keep testing and checking in production and let you know if I come across any other potential issues.

DiegoPino:
Resolved for now via https://github.com/esmero/format_strawberryfield/commit/1d8963d830b16d21880ea154c34ee5f67e32d05d
More code/issues will keep popping up. But fixes need to go in. Thanks

PR #444, opened 2024-04-22
ISSUE-442:Open API sanity + Controller cleanup

DiegoPino:
See #442 

More code to come @alliomeria


DiegoPino:
Merging for now. Of course we will find more issues bc this is complex science

Issue #445, opened 2024-04-23
Iabookreader with IIIF Manifest V3 fails to highlight when search term is passed via URL


DiegoPino:
# What?

New to me.

Was testing the IABookreader behavior today and noticed that, when IIIF Manifest V3 with page/individual hints is loaded the URL fragment (`#search/someword`) is ignored. This does not happen on a Manifest V2.

Not clear yet what/how this is happening but it might be related to how the default mode (the /mode/) is being added when a paging hint is found, not taking in account any previous passed /search. This is JS. 

@alliomeria will ask to replicate. Kinda a biggie bc people love their highlights

DiegoPino:
I figured out the issue. The URL processing Plugin is tied (race condition + bad code) to having Bookreader setup and init running almost at the same time/sequentially and blocking (there is a timed event listener). So when the manifest is being fetched async (so there is a wait) the URL Plugin that reads the fragment acts sooner + loses context (the "this" in JS in no longer in scope) and defaults to the saved cookie with the lasted/previous visit to a book.
The solution is to force the "defaults" when the Manifests ends loading by processing ourselves the fragment we saw while fetching the data, modifying it to match the desired/manifest level mode.
This opens the opportunity to make the fragment loading a function we manage (that works for this, heavily inspired in Bookreader) so I can call it also on OSD/Mirador and use that to allow links leading to search on both

Lots of work needed @alliomeria @patdunlavey but I will get it done before 2PM and merge

PR #446, opened 2024-04-24
ISSUE-445: Fragment not respected when async/manifest driven mode is present

DiegoPino:
# What?

See #445 

This is WIP. It does fix the actual problem but in some occasions (*when the results come back too fast?) the internal Bookreader logic resets the URL, so people could loose the ability of bookmarking. The code works but that is annoying. I need to find the culprit somewhere still

@alliomeria @patdunlavey


DiegoPino:
This will have to go into 1.5.0. Only I tested this, so not 100% accurate

Issue #447, opened 2024-05-20
Entity query access check needed in TwigExtension::entityIdsByLabel


patdunlavey:
Need to add `$query->accessCheck(FALSE);` right [here](https://github.com/esmero/format_strawberryfield/blob/1.4.0/src/TwigExtension.php#L145).

PR coming...

DiegoPino:
@patdunlavey cool. Thanks!

patdunlavey:
I almost always default to (FALSE) in these cases since that emulates the D9 behavior. Changing to (TRUE) could generate a regression, and should be more carefully evaluated. Do you object strongly to leaving it as (FALSE)?

DiegoPino:
@patdunlavey we don't use that extension or have used with any partners for that reason. I can see a point towards security being done there if used blindly with FALSE. Access to unpublished nodes, etc. 
For me the most concerning entities there are
"node" and "user"... so maybe adding access(TRUE) if one of those?
That all said, the function returns IDs... so basically once you try to load them/use them in any other way permissions will kick in anyways. 
What is your current use case/need of this one? Vocabularies?

DiegoPino:
Means I don't strongly object to FALSE, but want to be sure that if I have to make it TRUE in the future I will not break every site you have

patdunlavey:
We're using it in AMI set processing to convert strings to their corresponding taxonomy term ids. I *think* setting it to (TRUE) will be fine, but I don't want to be wrong and have to spend a lot of time sorting it out.

DiegoPino:
performance should not be an issue really, because the ::access() just adds under the hood a query condition, does not really load the entity but ok, let's keep it FALSE and I can write some if "node" alternatives later (taxonomies are kinda safe IMHO) but add (can you add a title to your pull too) a note to the Pull request so we can revisit this in the future? Thanks. Once you do that I can merge. If you need this backwards 1.3.0 please make a pull against 1.3.0 too, but given how different drupal 10.2 is to 10.1 I think 1.4.0 should be ok

patdunlavey:
I just did some double-checking the above-described AMI set use-case, and I'm satisfied that it's safe to set the accessCheck to (TRUE). We don't use this twig function anywhere else, but if we did, I think it's reasonable to impose the check in those cases. So I'll make that change in the PR. If that causes anybody difficulty, they can raise a new issue!

patdunlavey:
[...]

PR #448, opened 2024-05-20
Issue-447: missing accessCheck in TwigExtension::entityIdsByLabel

patdunlavey:
- Added entity query access check in TwigExtension::entityIdsByLabel

### DIFF
--- a/src/TwigExtension.php
+++ b/src/TwigExtension.php
@@ -142,7 +142,7 @@ class TwigExtension extends AbstractExtension {
         $query = \Drupal::entityTypeManager()
           ->getStorage($entity_type)
           ->getQuery()
-          ->accessCheck(FALSE);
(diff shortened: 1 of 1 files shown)


DiegoPino:
@patdunlavey since this is user exposed, probably ->accessCheck(TRUE) ?

DiegoPino (approved):

patdunlavey:
After discussion on the [issue](https://github.com/esmero/format_strawberryfield/issues/447), I agreed to set the accessCheck to TRUE. I also think we can leave 1.3.x as-is .

DiegoPino:
@patdunlavey if you have troubles with AMI imports (please check with background processing bc the user might be different) let me know and I make a quick power/move edit back to FALSE.

patdunlavey:
Thank you sir!

Issue #449, opened 2024-05-21
Better Metadata API Views argument/contextual filter and exposed filter handling


DiegoPino:
# What?

When the Open API accesses multiple Contextual filters, the order of those need to match exactly all existing ones. So that only happens if all are mapped/passed as arguments but not when one is present the other not.  This could be also avoided by having better Exposed filter matching but the idea is to allow both well (gosh)

@alliomeria I have to fix this during the morning. Will keep you updated. Thanks

alliomeria:
Thanks to you @DiegoPino !

DiegoPino:
Resolved via https://github.com/esmero/format_strawberryfield/commit/5e1a046c5437538c79262aae9b5b531d408efb7d

PR #450, opened 2024-05-21
ISSUE-449: Fix multiple Contextual filters, some "NULL" and allow Exposed filters. But avoid if already set same "key" via a Cont…

DiegoPino:
…extual one

Also, for contextual ones, use the "exception" value instead of NULL So we can avoid passing "NULL" and end being not validated because of that. $executable->exposed_data = $filters; was the trick ..

See #449 

@alliomeria this should do the trick with OAI and a single View dealing with both listing and getting a single record. Will report tomorrow with a live implementation and simple instructions for this (so you can also have this in your local for future needs/users)

### DIFF
--- a/src/Controller/MetadataAPIController.php
+++ b/src/Controller/MetadataAPIController.php
@@ -410,6 +410,7 @@ class MetadataAPIController extends ControllerBase
             }
             // Deal with Filters. Exclude values set already for the same field on Contextual ones
             // At least for now.
+            $filters = [];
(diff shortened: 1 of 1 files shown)


DiegoPino:
This works. Note. The Open API will not run any Views at all if none of the passed arguments is mapped to a Views argument or Filter. That is by design to avoid someone by mistake exposing their whole repository without a single filter (or though) via an API @alliomeria

Issue #453, opened 2024-06-26
Mirador Overrides 


alliomeria:
# What's happening:

When's using `JSON with Mirador Viewer configuration overrides`, some override parameters are not being respected (such as `{ "window": { "thumbnailNavigation": { "defaultPosition": "far-right" } } }`).

As seen during local tests by myself & @DiegoPino.

DiegoPino:
@alliomeria on it!

alliomeria:
Awesome, thanks @DiegoPino !

DiegoPino:
Solved via https://github.com/esmero/format_strawberryfield/commit/63570d92737066d91dca3b26b3265b141794ecf0

PR #454, opened 2024-06-26
ISSUE-453: Fixes Override Windows (array) loosing Manifest ID when overridden 

DiegoPino:
See #453 

To test, add this to the settings (Mirador settings are confusing.. this is how you put the thumbnail navigation for the first opened window to the right.
```JSON
{
"windows": [{
  "thumbnailNavigationPosition": "far-right"
}]
}
```

@alliomeria :eyes:


DiegoPino:
@alliomeria we could also totally remove from the per Window setting that defines where the Bar goes, and just leave the "default" you shared as driving factor. That would also make new windows also default to that position. We can talk tomorrow about that

DiegoPino:
@alliomeria I missed merging this for the release. Will merge now and will update composer locks tomorrow AM. Not a big deal

alliomeria:
Ok, all good, thank you @DiegoPino

Issue #457, opened 2024-08-01
Add middleware cache for Content Search API parsed IIIF Manifest.


DiegoPino:
# What?

All cool/nice/smooth when searching across 200 pages. Super slow on 767+ pages *also... that is a large manifest people. Why?

Solution is:

- Reuse the Metadata Exposed API caching mechanism, but instead of caching the actual IIIF Manifest (which will still require parsing via JMESPATH a 2.9 Mbytes (*for the 767 pages manifest expressed in this message) for every query YET again, we cache the already cleaned/processed JMESPATH results.

Caches will clear of course as always if the NODE or the Exposed endpoint or the IIIF Manifest itself changes.

@alliomeria, life is hard...

This connects with #456 of course. 

I almost feel tempted to allow people for these beast of ADOs to donwload the Manifest, upload back as file. Serve statically until the end of days (which is a manual version of #456 )

PR #458, opened 2024-08-04
ISSUE-457: Multiple bug fixes and improvements for Content Search API (Plain Text)

DiegoPino:
see #457 

I will make this 1.4.0 (since release happened but we are always allowed to make fixes until we start 1.5.0 branches & publish the release itself as a ZIP). 

This is not ready yet. I want to check what is happening with the SBF_ Aggregator processor and why it is triggered when the type of query has full processing ... Drupal gosh.


DiegoPino:
@alliomeria I will merge. More might come by Wednesday but this fixes several issues

Issue #459, opened 2024-08-08
Make ML similarity contextual filter and facets work


DiegoPino:
# What?

Facts: we refresh the View on a passed encoded Image + Node + File and fetch a vector from the NLP. We alter the query, we get the similarity results. This all happens via Ajax. We refresh the View and we refresh the Facets. Because these are different HTTP requests the facets don't have the `results` object anymore available, so they re-query (I dislike this a lot, so inefficient ..) and call the View again.
The view, if not cached, will re-query, but the original Contextual filter is no longer present in the request (bc we are filtering/making sure the requests are only done via Ajax/not public, etc) so the facets run without the KNN filter and that means in fact we always get the same facets back. As if KNN would have never run

Options we have:
- If Ajax enabled for the View we can "remember" in a session the Contextual filter parameters. ideally already the vector, so we do less backend processing. This is possible since All this ML tricks really can only run for logged in users. 
- Then on re-query we fetch the session. I can here check if the request was done via AJAX. so a "reload" of the page will never use the session. Good
- We need to be sure we actually clear the session on RESET.

DiegoPino:
Resolved via https://github.com/esmero/format_strawberryfield/commit/527d9cdd87caa17fdaa6f9e6a7b3b7489359b4c8

PR #461, opened 2024-08-22
ISSUE-459: Ajax/Facet/Views improvements  + ML similarity contextual filter and facets and 10.3 compatibility

DiegoPino:
See #459 

Also brings some Changes/fixes that happened for Drupal 11

https://git.drupalcode.org/project/drupal/-/commit/ef985e937c58257c4f73f034cfd9b828430f9ad6

### DIFF
--- a/js/mirador_strawberry.js
+++ b/js/mirador_strawberry.js
@@ -403,6 +403,25 @@
             $options.manifests = $manifests;
           }
 
+          const readFragmentSearch = function() {
+            const urlArray = window.location.hash.replace('#','').split('/');
+            const urlHash = {};
+            for (let i = 0; i < urlArray.length; i += 2) {
+              urlHash[urlArray[i]] = urlArray[i + 1];
+            }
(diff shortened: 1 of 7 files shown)


DiegoPino:
This starts cleaning the Ajax mess. Still missing a fix related to an issue that happens when you have multiple views in a single page, some have pagers and some have facets. A facet will only refresh the View it is attached to via AJAX, but won't wring its arguments to other Views that have pagers, which basically mean, when you page and click on something or filter, the URL itself of the page won't be passed around by the View to the next page, basically making "facet" selections to get stuck in space/time/state or being ignored. The solution is probably, on any View Update triggered by a Facet to also Check other/not to be updated Views and either
- Fully refresh (which might have the "feature" of allowing a single Exposed block facet (bc of the URL) drive many views?)
- detach and re-attach any pagers present for any view except the one just refreshed.

@alliomeria that is where I am now

Issue #479, opened 2024-11-12
Add file get content twig extension


DiegoPino:
# What?

## Why would you do that?

Imagine a Twig template that takes 5 minutes to process bc it uses thousands of results from Views and nested. Same as "this call could have been an email", "this twig template result could have been a static file" situation? Maybe attaching a JSONLD to the Object/ADO with a static result of that template is a better choice? But without adding more code to Mirador, etc. One could still serve the EXPOSED endpoint that would "if the JSONLD file is present and attached to the object then deliver its content wrapped around an API ... but if not, process the twig template as usual" type of scenario

Danke

PR #480, opened 2024-11-13
ISSUE-479: Adds the sbf_file_content extension

### LINKED ISSUE #483: Fix legacy Text option for mimetype/metadata display/api and exposed endpoints
# What?

Super old bug. The wrong mime type was being used for TEXT exposed data. Not a very common thing but needed to be fixed.

DiegoPino:
See #479 

works like this

```Twig
{{ sbf_file_content("uuid_of_the_node", "uuid_of_the_file", "thename.jsonld")|raw }}
````

Will return the content of a file if found/matches one of the mimetypes defined as valid for our metadata displays, now defined as constants at
```PHP
CONST ALLOWED_MIMETYPES = [
    'text/html' => 'HTML',
    'application/json' => 'JSON',
    'application/ld+json' => 'JSON-LD',
    'application/xml' => 'XML',
    'text/plain' => 'TEXT',
    'text/turtle' => 'RDF/TURTLE',
    'text/csv' => 'CSV',
  ];
```
an error message in case of exception or an empty string if the file does not exist, the user has no permission, is not of an allowed  type, etc (avoiding as much as possible errors)

Also fixes #483 

@alliomeria more tomorrow bc this solves VARIOUS use cases

### DIFF
--- a/format_strawberryfield.install
+++ b/format_strawberryfield.install
@@ -362,3 +362,23 @@ function format_strawberryfield_update_9001() {
     return 'Metadata API Configuration Entity already exists';
   }
 }
+
+
+/**
+ * Implements hook_update_N().
+ *
+ * Updates Mimetypes for text/plain
+ *
+ */
+function format_strawberryfield_update_9002() {
+  $definition_update_manager = \Drupal::entityDefinitionUpdateManager();
+  $field_storage_definition = $definition_update_manager->getFieldStorageDefinition('mimetype', 'metadatadisplay_entity');
+  $settings = $field_storage_definition->getSettings();
+  $settings['allowed_values'] = \Drupal\format_strawberryfield\Entity\MetadataDisplayEntity::ALLOWED_MIMETYPES;
+  $field_storage_definition->setSettings($settings);
+  $definition_update_manager->updateFieldStorageDefinition($field_storage_definition);
+}
+
+
+
+
--- a/format_strawberryfield.install
+++ b/format_strawberryfield.install
@@ -39,7 +39,7 @@ function format_strawberryfield_update_8001() {
         'application/json' => 'JSON',
         'application/ld+json' => 'JSON-LD',
         'application/xml' => 'XML',
-        'text/text' => 'TEXT',
+        'text/plain' => 'TEXT',
         'text/turtle' => 'RDF/TURTLE',
         'text/csv' => 'CSV',
       ],
--- a/format_strawberryfield.libraries.yml
+++ b/format_strawberryfield.libraries.yml
@@ -6,6 +6,7 @@ iiif_formatstrawberryfield_utils:
   dependencies:
     - core/jquery
     - core/drupal
+    - core/once
     - core/drupalSettings
     - format_strawberryfield/jmespath_strawberry
     - format_strawberryfield/svgpath_polyfull
--- a/format_strawberryfield.services.yml
+++ b/format_strawberryfield.services.yml
@@ -3,7 +3,7 @@ services:
     class: Drupal\format_strawberryfield\TwigExtension
     tags:
       - {name: twig.extension}
-    arguments: ['@renderer', '@plugin.manager.search_api.parse_mode']
+    arguments: ['@renderer', '@plugin.manager.search_api.parse_mode', '@request_stack', '@kernel']
   format_strawberryfield.deletetmpstorage_subscriber:
     class: Drupal\format_strawberryfield\EventSubscriber\formatStrawberryfieldDeleteTmpStorage
     tags:
--- a/js/iiif-archipelago-interactions_utils.js
+++ b/js/iiif-archipelago-interactions_utils.js
@@ -1,4 +1,4 @@
-(function ($, Drupal, drupalSettings, jmespath) {
+(function ($, Drupal, drupalSettings, jmespath, once) {
 
   'use strict';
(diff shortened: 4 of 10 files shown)


patdunlavey:
Hi Diego, I did some testing of [this commit](https://github.com/esmero/format_strawberryfield/blob/18b2f16a7c8cfe5396532f7089e27f60df630c0a/js/mirador_strawberry.js). Some things I noticed, though I'm not sure if they are *new* things (isn't that always the problem with requests to "test this"?!). In this table, "no" means I think this is or may be a problem, and "yes" means that I think it's correct.

| Test                                                                                                                                  | Mirador 3 from CDN                    | Mirador 4 from CDN                    | Custom Archipelago Mirador with Plugins (Mirador 4 from CDN selected) | Custom Archipelago Mirador with Plugins (Mirador 3 from CDN selected) |
| ------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------- | ------------------------------------- | --------------------------------------------------------------------- | --------------------------------------------------------------------- |
| Display mode defaults to "single" (1-up) ("no" means that it defaults to "book")                                                      | no                                    | no                                    | yes                                                                   | yes                                                                   |
| In "book" mode, page hash in the url is updated when the active page changes                                                          | no                                    | no                                    | no                                                                    | no                                                                    |
| In "single" display mode, page hash in the url is updated when the active page changes                                                | yes                                   | yes                                   | no                                                                    | no                                                                    |
| Manually adding "#page/N" to the url and pressing enter causes the browser page to reload with Mirador showing the correct page       | no (console shows a javascript error) | no (console shows a javascript error) | no (console shows a javascript error)                                 | no (console shows a javascript error)                                 |
| Pressing enter a second time reloads the object with the Mirador viewer showing the correct page (both single and book display modes) | yes                                   | yes                                   | yes                                                                   | yes                                                                   |

patdunlavey:
I'm sorry that I did not include testing search, since I don't have that set up in this project.

DiegoPino:
@patdunlavey sorry I don't understand the test matrix. I don't see the console errors, and also 1up/2up is not in what this delivers. Very confused. Could you provide more info? Thanks a lot

patdunlavey:
Thanks Diego. Here's the console error that I'm seeing (Linux Chrome) when I add/edit the `#page/NN` hash to the url and submit (press `<enter>`):
image
Pressing `<enter>` a second time triggers the page reload as I would expect.
I see the same error and behavior in Linux Firefox.

patdunlavey:
The other important thing that I'm seeing is that, when "Custom Archipelago Mirador with Plugins" is selected, I'm seeing Mirador always load in "book" (2-up) display mode. If that's not something you're seeing, that will be useful to know. (I can't think of anything I might be doing to inadvertently change the default display mode.)

DiegoPino:
For the 1up/2up thing I will say "we won't implement". There are many reasons why a user will not want an enduser/link to mess up with the display (selected) based on the actual IIIF Manifest. The manifest (and you opened an ISSUE which I solved) can and will define the display mode using the IIIF specs. And with the Mirador override JSON we could even disable (and some users have requested that) the Display Mode change on the UI, so allowing a link to trigger that bypassing the choices is a bad idea

patdunlavey:
I suspect form.js loads because we have the search form displayed on all pages (in the page header region).

DiegoPino:
With form.js, The issue there is an order of operations problem with Drupal. I can not tell Drupal in which order their internal libraries will be loaded. Happy though to not use IAbookreaders pattern at all, but then we (maybe you?) will have to make a wrapper code that uses the "new-valid element id" hash and transform/read on the fly at the IAbookreader level, bc we need to unify this approach
 (notice that iabookreader might be loading before form.js) 
 Or you can override form.js (will take a pull) to not blindly pass the hash into JQUERY ... to avoid the error

patdunlavey:
I tested the patch and, while it makes the console error go away, it seems to simply swallow it. The page still doesn't load.

patdunlavey:
Regarding the default 2-up thing I'm seeing, it seems to be consistent with `"behavior": ["paged"]` in the iiif v3 manifest. I tested changing it to "individuals", and it does indeed result in defaulting to single pages. 
But it's curious that... 
1. "paged" is the default in [archipelago-deployment](https://github.com/esmero/archipelago-deployment/blob/1.4.0/d8content/metadatadisplays/iiif_presentation_api_3.0_manifest.twig.html#L117), and...
2. Despite this, Mirador defaults to single page display when the "Custom Archipelago Mirador with Plugins" option is selected. 
Am I misunderstanding the purpose of the "behavior" attribute in the iiif manifest?

DiegoPino:
@patdunlavey not sure why that version of mirador (with the plugin) behaves like that.
Could you try commenting this here?
https://github.com/esmero/format_strawberryfield/blob/728535f0492e06e2853ba43a12eb0045203241e2/js/mirador_strawberry.js#L366-L388

Maybe the differentiated defaults we have for that JS are overriding the internal?

Let me know

DiegoPino:
>I tested the patch and, while it makes the console error go away, it seems to simply swallow it. The page still doesn't load.
Any messages in your logs?

DiegoPino:
@patdunlavey Will check if the way we alter the browser history is basically telling the caching of the browser that a reload is the same page or not. I can't reproduce the error you see and i get no errors in js at all, but will give this a try again

Issue #481, opened 2024-11-19
Remove hard-coded ado type field in ado type to view mode mapping form


patdunlavey:
I think this is really simple...
Change https://github.com/esmero/format_strawberryfield/blob/1.5.0/src/Form/ViewModeMappingSettingsForm.php#L224-L232
to...
```
    $field = $this->config('strawberryfield.archipelago_solr_settings.ado_type')->get('field') ?? 'type_1';
    $facets = [
      'type' => [
        'field' => $field,
        'limit' => 50,
        'operator' => 'AND',
        'min_count' => 1,
        'missing' => TRUE,
      ],
    ];
```

Is there any reason not to do this, @DiegoPino ?

DiegoPino:
not really. legacy decisions. GO for it

DiegoPino:
But it is not "hard" set ...
It is the default ... IF $this->config('strawberryfield.archipelago_solr_settings.ado_type')->get('field') is not get ...

patdunlavey:
I'm not sure I understand  your second comment. Something you think I should do different?

DiegoPino:
@patdunlavey please ignore my comment. I have too many open windows. Saw your suggestion and thought we were already doing it like that! (bc that was the original idea).

There are a few things we could improve for this form though (as you suggested in your pull) based on the fact that really the facet/selector is just a suggestion for already ingested types, and one basically could just use the "!" (the empty facet) to add "types" and mappings upfront before even ingesting anything (desired, specially if you are planning on controlling a bit what type of things people will eventually add), but this is already good.

DiegoPino:
Solved via https://github.com/esmero/format_strawberryfield/commit/95ae15f39af56108a11804cf0b6dd7a54b4cabfd

PR #482, opened 2024-11-19
ISSUE-481 -- Remove hard-coded ado type field in ado type to view mode mapping form

patdunlavey:
- On ado type to view mode mapping form, use the strawberryfield.archipelago_solr_settings.ado_type field configuration rather than assuming 'type_1' for the solr field that indexes the ado type.

See [ISSUE-481](https://github.com/esmero/format_strawberryfield/issues/481)

### DIFF
--- a/src/Form/ViewModeMappingSettingsForm.php
+++ b/src/Form/ViewModeMappingSettingsForm.php
@@ -221,9 +221,10 @@ class ViewModeMappingSettingsForm extends ConfigFormBase {
       ->getStorage('search_api_index')
       ->loadMultiple();
 
+    $field = $this->config('strawberryfield.archipelago_solr_settings.ado_type')->get('field') ?? 'type_1';
     $facets = [
       'type' => [
-        'field' => 'type_1',
+        'field' => $field,
         'limit' => 50,
         'operator' => 'AND',
         'min_count' => 1,
--- a/src/Form/ViewModeMappingSettingsForm.php
+++ b/src/Form/ViewModeMappingSettingsForm.php
@@ -4,6 +4,7 @@ namespace Drupal\format_strawberryfield\Form;
 
 use Drupal\Core\Form\ConfigFormBase;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Url;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\search_api\Entity\Index;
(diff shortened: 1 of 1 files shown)


patdunlavey:
@DiegoPino should we make this more robust? E.g. insist that valid archipelago_solr_settings configuration exist before we permit this form to submit?

DiegoPino:
@patdunlavey. The mapping (facets part) is a helper to allow people to "pick" types that are already ingested, but it is not authoritative. People should be able to plan upfront and add any type they want even before ingesting. The facet right now is also limiting in the sense that it assumes the "machine name for the field" is shared across Indexes and Servers. Which is most of the time not true. I am open to more validation, but not sure how you propose validation would happen.

DiegoPino (approved):
Looks good to me. Maybe we can add later on language saying that the list is what is ingested, but by picking the "!" you can later on edit the type and plan upfront? (see the language for that is already complex so asking)

src/Form/ViewModeMappingSettingsForm.php
```diff
@@ -55,7 +56,7 @@ public function buildForm(array $form, FormStateInterface $form_state) {
     $config = $this->config('format_strawberryfield.viewmodemapping_settings');
     $form['info'] = [
       '#markup' => $this->t(
-        'This Form allows you to map ADO (Archipelago Digital Object) "types" to existing Drupal View Mode Configurations.'
+        'This Form allows you to map ADO (Archipelago Digital Object) "types" to Drupal View Modes.'
```
DiegoPino, line 59: OK, I can see how this is shorter. Though the "existing" part is not wrong!

src/Form/ViewModeMappingSettingsForm.php
```diff
@@ -127,6 +128,22 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      '#collapsible' => FALSE,
+      '#collapsed' => FALSE,
+      '#tree' => TRUE,
+      'notes_list' => [
+        '#theme' => 'item_list',
+        '#items' => [
+          $this->t('"The ADO Type" refers to the solr indexed value as configured in the Archipelago @solr_settings_link form. Note that it could be referring to any "type" value nested at any level in your JSON metadata.',
+            ['@solr_settings_link' => \Drupal::service('link_generator')->generate('Important Solr Settings', Url::fromRoute('strawberryfield.important_solr_settings_form'))]),
+          $this->t('You may select "!" if you need to map an ADO JSON "type" value to a View Mode where that type is not yet available. After creating the mapping, you can edit the ADO Type directly.'),
+          $this->t('If more than one view mode mapping is enabled for a given "type" value, the <strong>first</strong> will prevail.'),
```
DiegoPino, line 143: This all reads very well @patdunlavey ! Thanks so much

src/Form/ViewModeMappingSettingsForm.php
patdunlavey: Not wrong, but not necessary either. Is it a view mode if it doesn't exist? :grinning:

DiegoPino:
@patdunlavey Your extra notes really add to the whole experience. Thanks so much. I added a comment about the change in wording to " 'This Form allows you to map ADO (Archipelago Digital Object) "types" to Drupal View Modes.'".  Just curiosity, why do you feel "existing" was wrong? misleading? (or just redundant?)

patdunlavey:
Not wrong, just redundant.

DiegoPino:
PERFECT. Merging then. Thanks!

PR #495, opened 2025-03-03
Adds all settings of Pannellum Viewer to iiif-pannellum_strawberry.js

roromedia:
This PR adds all available settings of the Pannellum Viewer available here: https://pannellum.org/documentation/reference/

The correct as:formatter JSON representation to make new settings work is like this:

 "as:formatter": {
            "urn:uuid:b999aec3-dd4d-4894-af5e-7f8f0b651f1e": {
                "settings": {
                    "mouseZoom": false
                }
            }
        }
    }
    
    The urn-value is your media key of the panorama item.


DiegoPino:
@roromedia Thanks! I need to do some testing here/small refactoring. Specifically bc this here:

https://github.com/esmero/format_strawberryfield/pull/495/files#diff-651210e04cdae7c4b67ae6efb89dcc4cf563df86c160a76f675fbd2eab2c5985R175-R189

Comes too late into play. basically at that point as:formatter can't override the set $mouseZoom etc variables for normal panoramas.

Will take me a few hours testing but will make a pull against your branch so you can confirm my changes don't break your own needs. Thanks again!

DiegoPino:
supper seeded by #499

PR #499, opened 2025-03-07
ISSUE-498: Allow Panorama tour and single Panoramas to overriden on all properties

DiegoPino:
@roromedia this includes your commits too. I hope this aligns with your needs and also the other override options.

Using this you can now avoid the more complex override that used the media key, and go for overrides at the formatter/settings level using the new JS override

like 
```
{
"mouseZoom": false
}
```

AND/Or at the Panorama Tour level like:

```
"ap:viewerhints": {
            "author": "RoRO"
        }
    },
```

AND/OR at the individual Panoramas referenced from the tour (so at the Panorama ADO level/raw JSON) via 

```
"ap:viewerhints": {
            "mouseZoom": true
        }
    },
```

And of course using as:formatter instead of "ap:viewerhints" and/or a $mediakey if you prefer that instead (which is not standard but I left it as a config option since this JS/PHP was the first one we extended before we made "ap:viewerhints" a global Viewer Override option)

Please test and let me know

This pull overrides #495 but includes your commits and also preserves your authorship/attribution.

Thanks

### DIFF
--- a/js/iiif-pannellum_strawberry.js
+++ b/js/iiif-pannellum_strawberry.js
@@ -146,9 +146,47 @@
           const $vOffset = drupalSettings.format_strawberryfield.pannellum[element_id].settings?.vOffset;
           const $maxPitch = drupalSettings.format_strawberryfield.pannellum[element_id].settings?.maxPitch;
           const $minPitch = drupalSettings.format_strawberryfield.pannellum[element_id].settings?.minPitch;
+          
+          // Additional Pannellum settings from documentation
+          const $hfov = drupalSettings.format_strawberryfield.pannellum[element_id].settings?.hfov;
+          const $minHfov = drupalSettings.format_strawberryfield.pannellum[element_id].settings?.minHfov;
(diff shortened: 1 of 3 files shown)


DiegoPino:
@roromedia to ease your testing I will merge this but won't close the ticket. If I made a mistake or something is missing I can make another pull

js/iiif-pannellum_strawberry.js
```diff
@@ -138,107 +139,149 @@
...
+              "hfov": 100,
+              "minHfov": 50,
+              "maxHfov": 120,
+              "pitch": 0,
+              "yaw": 0,
+              "horizonPitch": 0,
+              "horizonRoll": 0,
+              "compass": false,
+              "northOffset": 0,
+              "preview": "",
```
roromedia, line 166: Here the default value should be $preview || '' otherwise the poster image won't be used.

roromedia (commented):
The we got a regression as it doesn't load the poster image anymore.

Issue #501, opened 2025-03-12
Flavour (presence at multiple levels) Block Condition


DiegoPino:
# What?

Sometimes/somedays you need to position a block based on the knowledge that your ADO, or a Child of an ADO, of a GrandChild of an ADO has a particular Strawberry Flavor (Flavour as in British English)  extraction. E.g, only Show Text Transcripts in a Block (and the block itself) if the Object or one if its Children has that (mixes well with a IIIF Manifest that uses Annotation Pages with attached TEXT).

Bc a block has no idea what manifest is loaded, doing a Special condition block that (based on the vast user knowledge of their data and files and manifests and viewers) allows the user to hide/show based on that is useful.  

That is the ticket @alliomeria

DiegoPino:
Resolved via https://github.com/esmero/format_strawberryfield/commit/8b474e1a2e6aac1064b3c8dce03b933078f0578f

PR #502, opened 2025-03-12
ISSUE-501: Adds the Flavour Condition for Blocks

DiegoPino:
See #501 

Simple. Well written. perfect British (with fancy accent too). Does the job and works,

Looks like this

image

depends on https://github.com/esmero/strawberryfield/pull/351


DiegoPino:
*the screenshot also shows the File Condition.. that is another pull, a next week problem, kay?*

Issue #511, opened 2025-04-22
Add Metadata (or any other Solr field) search option to IIIF Content Search API


DiegoPino:
# What?

As discussed yesterday @alliomeria, users might want to find Canvases based on metadata. I thought about a Template/Metadata based Strawberry Runners Plugin as the only solution, but I think allowing users to Pick from Full Text fields (Multiple selections) might be easier. And the match ADO (might be multiple canvases, or a single) in a manifest to Matched "fields".

That approach would also allow in the future to have that special Strawberry Runners, but as a side effect, will allow us to start using this without indexing/re-processing anything as of now (now == when I code this)

let me know your thoughts/concerns, if any

DiegoPino:
Resolved via https://github.com/esmero/format_strawberryfield/commit/84284bc5f6365a1883db2a1aa7bc58f760e49938

PR #478, opened 2024-10-24
ISSUE-443: Better Date Range Facets  + Core Ajax improvements + New conditionals

DiegoPino:
See #443 
Complete Refactor. Better JS driven, better PHP, safer, more options. Overrides query processor and uses actual backend ranges with granularity to process Sol JSON API Facet & stats plus graph via Chart.js

Also corrects lots of issues with Ajax/Facets (overrides), new Mirador versions and ML related JS / Strawberry Runners data

Also, IIIF Content Search API for metadata. Gosh. This is a lot (sorry Dev deities .. this is the new 2025-me)

Includes #480 and #483

PR #512, opened 2025-04-25
ISSUE-511: IIIF Content Search API with Metadata/per canvas search 

DiegoPino:
See #511 

Because I have too many pending changes on #478 I decide to cherry pick into this new one. Need to test though if that did not  break anything, so please don't merge yet


DiegoPino:
Ok @alliomeria this works. Tested, nothing breaks. Will merge. If I find an issue, or needs improvement I can make a new pull


What is really different now is the IIIF Settings form
- No longer required fields if IIIF Content Search API is not enabled (finally!)
- New checkboxes to enable the Metadata search. Will request Wording improvements (different pull will follow if so)
- We will have to update documentation for 1.5.0 for this
image

Issue #524, opened 2025-08-21
Improve (one more time/not the last time) the Date Range Facets


DiegoPino:
# What?

BCE dates. Strange Math logic. Native Date range fields v/s date Fields (like both return different data structures), more options for the widgets and even more stricter validation all around.

Connects with https://github.com/esmero/strawberryfield/issues/358

I have code/I will push that code

@alliomeria @patdunlavey

DiegoPino:
Resolved via https://github.com/esmero/format_strawberryfield/commit/ace368e52f98c9af3209d50011ccebda385b9b7b
Can always be better/more precise, guess people's date wishes and time span considerations. But performance is also important. Sooooo... Happy with this.

PR #525, opened 2025-08-21
ISSUE-524: better native solr date range handling, better date math/timezone and also stricter checks on User input

DiegoPino:
# What?

See #524 and follow us on slack for more "dating" tips 😮‍💨 

Still missing here:
- ~add BCE/AC select box to the year/full date entry boxes in the widget~. Done differently. Full date input hides if one of the ranges is negative. Less UI for now is good.
- ~Totally remake the JS side of things date parsing bc the timestamps for BCE are totally wrong (no clue yet how)~ DONE
- ~Check/double check all Widget settings we have. See if all are respected.~ DONE (the old range widget was not working. mea culpa. Fixed)
- ~Incorporate a few shared ideas in slack like  allow "extra slack/+-gap" to any start/end when faceting via JSON on the backend, that way even if the query is 2021-2025(my most complex years)~. So far the solution is approximate but good enough. The last gap now uses the modulus of the GAP (bc it will not be exact) and we pass that data between query and widget back and forth. I can get an histogram that returns 2018-2028 (with 2028  maybe a good year?)

In the case of a single range (1899-01-01 to December 31st of the same year) the histogram turns into a single bar, If not it becomes invisible. 

Will follow up with more pulls for that tomorrow

### DIFF
--- a/modules/format_strawberryfield_facets/js/slider.js
+++ b/modules/format_strawberryfield_facets/js/slider.js
@@ -35,8 +35,12 @@
       let ctx = element.getContext('2d');
       let chart_data  = JSON.parse(settings.chart_data);
       let chart_labels = JSON.parse(settings.chart_labels);
+      let chart_type = "line";
+      if (chart_data.length == 1) {
+        chart_type = "bar"
+      }
       let histogram = new Chart(ctx, {
-        type: 'line',
+        type: chart_type,
         data: {
           labels: chart_labels,
           datasets: [
--- a/modules/format_strawberryfield_facets/js/slider.js
+++ b/modules/format_strawberryfield_facets/js/slider.js
@@ -48,7 +48,7 @@
               fill: true,
               stepped: true,
               showLine: true,
-              tension: 0.5
+              tension: 0.4
             },
           ]
         },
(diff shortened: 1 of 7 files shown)


patdunlavey:
@DiegoPino I'm going to be out of town until Monday. I'll review then if you still need.

DiegoPino:
@patdunlavey enjoy the weekend. Let's do a rain check on Monday. Thanks

DiegoPino:
@patdunlavey. I believe I have to keep further refining this pull. It does work (at least better than before) and native Solr date ranges work too, but I have the feeling the cut-off/end range is being reduced more than I would like and there might be a floating point (or any other math) issues lingering still or it might be the cut-off settings of the ranged faceting sent to Solr too. Debugging is required to be sure. Also, if you test please fetch a fresh strawberryfield 1.5.0 bc you might need it if you decide to go for BCE dates.

I can come back to this on Friday (have to travel tomorrow). Thanks a lot

DiegoPino:
@patdunlavey I will wait until the end of the day for comments on this. If not I will merge because as it is right now it is a real improvement over the previous logic. Thanks!

Issue #526, opened 2025-08-22
Make IP embargo bypass == TRUE (global config evaluation instead of local) more flexible to other "true" values


DiegoPino:
# What?

Allow that user configured key (normally `ip_embargo_bypass` key also evaluate as "GLOBAL" if

== TRUE (strict boolean as we do now)
== "1" (the number one as a string)
== 1 (the number one as an integer)

Reason is sometimes a webform element used for this might be setup to not explicitly set "TRUE" but either of the other ones. Reading, TRUE, "1" or 1 back on those elements will still work because of how Checkbox data is cast.

DiegoPino:
resolved via https://github.com/esmero/format_strawberryfield/commit/357924f6bdae900ac562840da859222c16babb5f

PR #527, opened 2025-08-22
ISSUE-526: Allows TRUE, "1" and 1 as valid global IP embargo flags via metadata

DiegoPino:
See #526 

@alliomeria this is what we have running now. Tested. I don't see any issues but please check one more time and merge (today hopefully) if you can. We do might need a DOCS update to state that "1" and 1 are allowed?

Thanks


alliomeria:
Looks + works great, @DiegoPino! Merging. Also: noted about a future docs update about the 1 + "1" clarification too.

Issue #530, opened 2025-09-17
Don't do metadata search via IIIF Content Search if no "query" term is passed.


DiegoPino:
# What ?

The "no highlights" because no term is passed is wrongly calculated for the new Metadata Discovery functionality of the IIIF Search endpoints, which also lead me to figure out that I should not even do a query if term is empty (includes multiple spaces)

@patdunlavey @alliomeria and I have a fix. Pulling and merging in 5. Thanks

DiegoPino:
Note: using mirador this is never hit. But direct access to the API without a viewer will show it. So basically a non reported error  because unknown to fail in the wild

DiegoPino:
Resolved via https://github.com/esmero/format_strawberryfield/commit/8fb76ea5713ea4c4591965b116ffdd75f5b19fa9

PR #531, opened 2025-09-17
ISSUE-530: Fixes metadata custom search on IIIF Content Search API under a "no query" situation

DiegoPino:
See #530. The fix is actually super short (I missed a warning under PHP 8.3.. sorry) but the changes look big because of indents.


DiegoPino:
@alliomeria tested. Just a minor thing but I had a bit of technical debt there!

Issue #532, opened 2025-09-18
Metadata Display Download Form


DiegoPino:
# What? 

I am too old to json-encode/copy paste bodies and clean up manual JSON-API GET calls to make them JSON-API-PATCH/PUT compatible. Translates to "I am too old in general:

Similar to https://github.com/esmero/format_strawberryfield/issues/423
But (baby steps) per Metadata Display first

Provide both twig.html and JSON API ready formats.

DiegoPino:
Resolved via https://github.com/esmero/format_strawberryfield/commit/196ae346cf59b904be4de1f56a426cf531128af5
A "Download all", as a zip will follow, but not today!

PR #533, opened 2025-09-18
ISSUE-532. Metadata display Download

DiegoPino:
See #532


DiegoPino:
This looks reasonable good and will merge. Note: You need to have permission to edit a twig template to download it. Just being overly secure here bc templates have literally a lot of logic

Issue #538, opened 2025-10-22
Date range facet widgets fail form label and contrast accessibility checks


patdunlavey:
We are seeing form label errors and low contrast errors on the "Format Strawberryfield Date Range slider" (sbf_range_date_slider):
<img width="869" height="586" alt="Image" src="https://github.com/user-attachments/assets/eef12a3f-33e6-495e-a7d0-695fe9ca99b2" />

Presumably we are seeing similar errors on the "Format Strawberryfield Date Range Picker (No Slider)" (sbf_date_range) as well.

The form label errors - an absence of a "for" attribute on the label elements - is caused by the form render arrays not including an ['#id'] value for the labeled form elements. (['#attributes']['id'] does not suffice, apparently).

The low contrast errors come from jquery-ui-slider-pips. That library is no longer being maintained, so the simplest solution, I think, is to provide overriding css for those slider labels.

I have patched the format_strawberryfield module (1.5.x-dev) to fix the form label errors and can offer a PR for that. 

I fixed the low contrast errors in my theme's css, but can re-roll that into a PR for the format_strawberryfield module as well, if you like! (format_strawberryfield/modules/format_strawberryfield_facets/css/slider.css)

DiegoPino:
@patdunlavey thanks for bringing these accessibility issues! Few questions

are the  "for" attribute on the label elements a responsibility of the underlaying theme (templates) handling each element or on us? If on us happy to provide solutions. The Fixed IDs (#id) might be an issue at the theme level or render array (solvable). There is a trend of moving out of fixed IDS (bc of Ajax) into data attributes.

Low contrast can be solved, true, via CSS overrides and so happy and grateful to get contributions for that from you (please against 1.6.0)! Pull away!

Happy to code for a different slider JS, if you have a proposal, I could add an "config option" that works around that.

patdunlavey:
@DiegoPino I don't know how other themes work, but the bootstrap_barrio (base) theme expects the ['#id'] value as I described. Its form-element.html.twig file utilizes `{{ label }}` in some places, so I believe that fixing it at the theme level would entail render array bubbling nonsense. I spent a while trying to do exactly that, and gave up because I couldn't work out how to fix it in either a preprocess function or in a template (without completely hacking the template). Maybe I missed some easy trick? 

Will using `Html::getUniqueId()` keep us ajax-safe?

[...]

PR #539, opened 2025-10-22
ISSUE-538--date-range-widget-accx ---> 1.6.0

patdunlavey:
Proposing accessibility fixes for date range slider facet widget and date range facet widget:
See #538 

1. Added ['#id'] values to each labeled form element so that the labels will get rendered with a 'for="\<the id\>". This associates the labels with the form elements for screen readers. We use Html::getUniqueId() as the basis for these ids. Hopefully this keeps them ajax safe?
2. Altered the color values of jquery-ui-slider-pips labels to meet WCAG AA contrast requirements.

### DIFF
--- a/modules/format_strawberryfield_facets/css/slider.css
+++ b/modules/format_strawberryfield_facets/css/slider.css
@@ -17,3 +17,42 @@
   color: black;
   color: var(--bb-primary-dark, var(--bs-body-color, black)) !important;
 }
+
+/**
+ * jQuery UI Slider Pips - Accessibility Contrast Fixes
+ * Overrides for jquery-ui-slider-pips library to meet WCAG AA contrast requirements (4.5:1 for normal text)
+ */
+.ui-slider-pips .ui-slider-pip {
+  color: #4a4a4a; /* 9.28:1 contrast ratio on white */
+}
+
+.ui-slider-pips .ui-slider-line {
+  background: #4a4a4a; /* 9.28:1 contrast ratio on white */
+}
+
+.ui-slider-pips [class*=ui-slider-pip-initial] {
+  font-weight: bold;
+  color: #0d8a58; /* 4.69:1 contrast ratio on white */
+}
+
+.ui-slider-pips .ui-slider-pip-initial-2 {
+  color: #106b96; /* 5.26:1 contrast ratio on white */
+}
+
+.ui-slider-pips [class*=ui-slider-pip-selected] {
+  font-weight: bold;
+  color: #c05500; /* 5.74:1 contrast ratio on white */
+}
+
+.ui-slider-pips .ui-slider-pip-inrange {
+  color: #000000; /* 21:1 contrast ratio on white */
+}
+
+.ui-slider-pips .ui-slider-pip-selected-2 {
+  color: #cc0000; /* 5.3:1 contrast ratio on white */
+}
+
+.ui-slider-pips [class*=ui-slider-pip-selected] .ui-slider-line,
+.ui-slider-pips .ui-slider-pip-inrange .ui-slider-line {
+  background: #000000; /* 21:1 contrast ratio on white */
+}
--- a/modules/format_strawberryfield_facets/css/slider.css
+++ b/modules/format_strawberryfield_facets/css/slider.css
@@ -22,37 +22,37 @@
  * jQuery UI Slider Pips - Accessibility Contrast Fixes
  * Overrides for jquery-ui-slider-pips library to meet WCAG AA contrast requirements (4.5:1 for normal text)
  */
-.ui-slider-pips .ui-slider-pip {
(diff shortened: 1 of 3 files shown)


modules/format_strawberryfield_facets/css/slider.css
```diff
@@ -16,4 +16,43 @@
 .facets-widget-sbf_range_date_slider .form-control {
   color: black;
   color: var(--bb-primary-dark, var(--bs-body-color, black)) !important;
-}
\ No newline at end of file
+}
```
DiegoPino, line 19: @patdunlavey this all looks great. Just to avoid overriding other people's specific choices on the slider, could you make the CSS overrides target .facets-widget(something-other-more specific class) .ui-slider-pips etc? That way other users with other needs (or other pips usages) don't get the overriden, but also, the inverse, others don't override ours? Thanks a lot

DiegoPino (changes requested):
Small change. add an extra refining class prior to .ui-slider-pips, so other themes and modules don't compete with these settings so easily

modules/format_strawberryfield_facets/css/slider.css
patdunlavey: I found myself arguing this both ways, but I decided to agree with you that we should make the selectors target `.sbf-date-facet-slider.ui-slider-pips`. (My counter-argument was that we're correcting a general accx problem, so our fix should target the general problem. But I decided that doesn't fully hold water. We're only loading our css for our widget, so we aren't solving the problem generally anyway.) 
Updated PR coming...

patdunlavey: New commit added: https://github.com/esmero/format_strawberryfield/pull/539/commits/2c44df960df250fe548c1b0caaa5edd462ccc2f8

modules/format_strawberryfield_facets/src/Plugin/facets/widget/DateRangeSliderWidget.php
```diff
@@ -33,25 +33,28 @@ public function build(FacetInterface $facet) {
...
 
     if (($this->getConfiguration()['allow_full_entry'] ?? FALSE) && ($this->getConfiguration()['allow_year_entry'] ?? FALSE) && !$is_bce) {
       $build['#items']['manual_input']['select_input'] = [
           '#type' => 'checkbox',
+          '#id' => $id . '-manual-input-fulldate',
           '#title' => t('Full Date entry'),
           '#default_value' => FALSE,
           '#attributes' => [
-          'data-date-entry-selector' => $id.'-fulldate'
+          'data-date-entry-selector' => $id . '-manual-input-fulldate'
```
DiegoPino, line 57: @patdunlavey another one I just noticed. We use the states api to hide/show things based on this selector. By moving it from -fulldate are we breaking something somewhere else? 
Could you please test this here?


https://github.com/esmero/format_strawberryfield/blob/a1da3d9b48f785cd1d82df5abbff984a3f2340ba/modules/format_strawberryfield_facets/src/Plugin/facets/widget/DateRangeSliderWidget.php#L132-L153

DiegoPino: Maybe bring also the change into that, replacing with selector to $id . '-manual-input-fulldate' to match the fact that $id does not longer contain -manual-input since it is being used as base for other `#ids`?

patdunlavey: You're right - I broke it! Good catch!! I'll get that fix done.

DiegoPino: Thanks!

patdunlavey: Hmmm... that was an easy fix, but I'm finding that the full date filtering isn't working. I'm not sure if this is new. In our implementation, we don't have full date filtering enabled, so I haven't tested it before. Can you confirm that this is a new problem @DiegoPino ?

DiegoPino: @patdunlavey true. Can confirm. Mine was disabled bc we have dates that are BCE. I enabled it by setting it fixed to 0 as min and it is not working. Probably I mistake I made on the last date check JS update. Will fix in another commit, unrelated to this.

DiegoPino (approved):
@patdunlavey approving. The Full date submit logic (JS) is unrelated to this pull and I opened another issue for that. Your changes look good. Will merge!

DiegoPino:
@patdunlavey please commit your #state fix before I merge? I did not see that one come in!

patdunlavey:
Oh, yeah! Ok, done.

Issue #540, opened 2025-10-22
Date Range Full date entry submit not working after post release 1.5.0 JS updates


DiegoPino:
# What?

Reported by @patdunlavey . The submit/read from the field logic is not working anymore. A JS. Will patch back 1.5.0 and 1.6.0 once I figure it out.

PR #541, opened 2025-10-22
ISSUE-540: 1.5.0 backwards fix

DiegoPino:
See #540

### DIFF
--- a/modules/format_strawberryfield_facets/js/slider.js
+++ b/modules/format_strawberryfield_facets/js/slider.js
@@ -204,7 +204,7 @@
               max = ui_max = date_from_input.getFullYear();
               let month = date_from_input.getMonth();
               let day = date_from_input.getDay();
-              max_timestamp = toTimestamp(min, month, day, 0, 0, 0, 0);
+              max_timestamp = toTimestamp(max, month, day, 0, 0, 0, 0);
             }
           }
         }
--- a/modules/format_strawberryfield_facets/js/slider.js
+++ b/modules/format_strawberryfield_facets/js/slider.js
@@ -192,7 +192,7 @@
         else if (e.target.type == "date") {
           if (e.target.dataset?.type == "date-range-min") {
             let date_from_input = new Date(e.target.value);
-            if (typeof date_from_input == Date) {
+            if (date_from_input instanceof Date) {
               min = ui_min = date_from_input.getFullYear();
               let month = date_from_input.getMonth();
               let day = date_from_input.getDay();
@@ -200,7 +200,7 @@
             }
           } else {
             let date_from_input = new Date(e.target.value);
-            if (typeof date_from_input == Date) {
+            if (date_from_input instanceof Date) {
               max = ui_max = date_from_input.getFullYear();
               let month = date_from_input.getMonth();
               let day = date_from_input.getDay();


patdunlavey (changes requested):
I tested out this change, and I'm seeing that the min date submitted does not match what I selected from the calendar widget.

Screenshot of my selection, prior to submit:
image

After submit:
image

Not sure if this is in-scope for this issue though! I'm going to mark it as "request changes" unless you tell me that it's out of scope for this issue.

patdunlavey:
Also noticed that the max date changed too. What we see displayed in the facet and in the search results matches the min and max date timestamps in the url. So it seems that the problem is in the submit side of things.

DiegoPino:
@patdunlavey, the updated value back is because the end date from the results is lower. 

The submit (JS code) only does a Date to unixtimestamp transformation. If the results coming back are not exactly the input it is because of timezone? Or actual matches for that timezone being different?. If that needs fixing then it will have to be a different issue. You OK with that?

DiegoPino:
If you have any suggestions on how to approach the issue (you said the timestamp in the URL is correct?) let me know.
The function used by both the slider and the year and the full date entry is the same, maybe I have a scope issue. But I would like to debug that differently (when/if I understand the issue)

  function toTimestamp(year, month, day, hour = 0, minute = 0, second = 0, millisecond = 0) {
        // Date.parse works only on safari for BCE!
        // New approach. Date Constructor
        let date_from_parts = new Date();
        date_from_parts.setFullYear(parseInt(year));
        date_from_parts.setMonth(parseInt(month)-1);
        date_from_parts.setDate(parseInt(day));
        date_from_parts.setHours(hour,minute,second,millisecond);
        const datum = date_from_parts.getTime();
        if (Number.isNaN(datum)) {
          return Math.round(Date.now() / 1000);
        }
        return Math.round(datum / 1000);
      }

DiegoPino:
Pat. in other words

Input *you (max and min) lead to different max and min in the faceted results. The facet summary is based on your input, slider, year input and full date input are based on what we are getting on the backend, and might not match what you "inputed". How fine grained those results are (GAP between each, specially if you are using date ranges) will depend on your max facet count *facet setting because Ranges are all "grouped" bucketed results. I think the logic is Ok. But I can be convinced it is not?

DiegoPino:
@patdunlavey @alliomeria ok... good I never dismiss any reviews and I mistrust myself. There was INDEED an issue in the **max** calculation. That should be OK now. There will be (or could be) still a precision / requests v/s real results mismatch, and the results will drive the "updated" values in the widget not the submitted values. That is expected and needed since the slider range needs to match also the possible manual inputs. let me know if that fix makes things less "random" now. Thanks

patdunlavey:
Sorry - had to step out for an hour or so! You asked if the timestamp in the url is correct, and what I meant is that the timestamp in the url was _incorrect_ as regards the values in the form prior to submitting, but that it matches the dates displayed in the date fields after the page reloads. In my example, the date values in the form when I submitted were `11/01/1979 to 12/31/2020`, but when the page reloaded, the values displayed in the date widget were `09/03/1979 to 12/31/2019`. The timestamps in the url after the page reloaded do correspond with the dates displayed in the widget after that reload.

I tested it more after your [fix to max_timestamp](https://github.com/esmero/format_strawberryfield/pull/542/files#diff-945ce3d155a423d5984a3cad6ae67cdb84494762f3205b740027c79610762751L207-R207). What I think is happening is a function of the facet granularity. The user inputted date values still don't match the result date values. It looks like the result date values are set by the available min and max dates based on the granularity. I _think_ that makes sense, though it's very non-intuitive. I think users expect the date that they enter to be echoed back to them in the results. That only works if they happen to select dates that line up with the granularity of the facet. I wonder if it would be possible and make sense to hide this conversion to the granular date buckets? In other words, the search always displays the date values that the user enters, even though the actual facet filtering is converting to the nearest (or rounded up/down) date bucket?

DiegoPino:
@patdunlavey ok, I understand (and have thought about it while I was coding it) that having the range reduce itself after input submissions (change the input to the new max/min) might be confusing to -some- users but, reason I made it, is bc it is also safer (basically blocks / at least partially - except when the range has "holes" - the user from selecting ranges after the initial submit that have no data... but the Histogram should inform users anyways?). So... I will open an issue and allow that to be a facet widget option for 1.6.0. In the meantime, if the actual (fixed?) "logic" of timestamp submissions I now Ok, could you please approve this one and the 1.6.0 one too? If not, let me know where the inconsistency is. Thanks a lot!

patdunlavey (approved):
Having an option in the future to switch the date filter mode would be ideal! I'm satisfied with your explanation and hereby approve this pull request!
Thanks!!

patdunlavey (approved):
Good to go. (Sorry I forgot we have two PRs!)

DiegoPino:
Muchas gracias @patdunlavey . Merging!

Issue #574, opened 2026-02-11
Expand leaflet_strawberry.js to Zoom out for Polygons and Multipolygons


markpbaggett:
Hey all, 

I'm working on a giant map collection and we have geojson polygons and multipolygons that we want to show instead of markers and marker cluster groups.  The good thing is that this all currently works out of the box.  The only problem is that the Zoom level logic is missing for polygons.

We're joining AllMaps so we expect to have lots of georeferenced content like this. Can we add a tiny bit of code so zoom level works:

```
    geojsonLayer.on('data:loaded', function () {
      if (!cluster_added) {
        markers.addLayer(geojsonLayer);
        map.addLayer(markers);
        // Use geojsonLayer bounds so that Zoom level works for both points and polygons
        if (geojsonLayer.getBounds().isValid()) {
          map.fitBounds(geojsonLayer.getBounds());
        }
      cluster_added = true;
      }
    });
```

DiegoPino:
@markpbaggett sounds like a good use case. Want to give a pull a try? 

Should  we add a config option in case someone wants to fix the zoom level to a specific value instead of going for the (better) bounding box?

markpbaggett:
Working on something right now and will open a pull request in a sec that we can further discuss.

DiegoPino:
Cool and thanks! Take your time, I am closing the bakery for the day here. Will check back tomorrow morning!

PR #575, opened 2026-02-11
Address Issue-574.

markpbaggett:
## What Does This Do

This attempts to address [ISSUE-574](https://github.com/esmero/format_strawberryfield/issues/574), but my logic is a little different than I originally suggested. Instead:

1. If we only have 1 point, setView at configured zoom like before, because a single point has no meaningful extent to fit into a view.
2. If we encounter anything else: a polygon, a line, multipolygons, or multiple points and need a MarkerClusterGroup, we use fitBounds to fit all geometry.

In other words:

1. We now calculated bounds via `geojsonLayer` instead of marker cluster groups. This means polygons, multipolygons, and lines would contribute to the viewport calculation as well.
2. Single features now only apply to actual points and we detect by checking if NE and SW corners of the bounds are identical, i.e. zero area. 

Also, I added an `isValid()` guard so the map doesn't attempt to zoom on invalid data.

Here is a single marker:
Screenshot 2026-02-11 at 5 14 48 PM

Here is a rushed together polygon:
Screenshot 2026-02-11 at 5 15 27 PM


js/leaflet_strawberry.js
```diff
@@ -107,12 +107,19 @@
             geojsonLayer.on('data:loaded', function () {
               if (!cluster_added) {
                 markers.addLayer(geojsonLayer);
-                if (geojsonLayer.getLayers().length > 1) {
-                  map.addLayer(markers).fitBounds(markers.getBounds());
-                } else {
-                  map.addLayer(markers).setView(markers.getBounds().getCenter(), $initialzoom);
+                map.addLayer(markers);
```
DiegoPino, line 110: Hi. Small performance idea here

`geojsonLayer.getBounds()` can be expensive if we have many points and polygons and the method/class does not cache calculated version. So every call re-calculates. You have two calls. So I would suggest you do the following instead

```JS
const bounds = geojsonLayer.getBounds();
if (bounds.isValid()) {
```

then you don't need to re-call get bounds again or assign to a `var` `bounds` at all. 


One feature (will an open issue, this one can go in anyways) I would love to preserve via a formatter setting/override is the idea of having still a way of setting a fixed zoom, or at least one that respects one issue we have found. Some tile sources don't have enough detail for certain zoom levels. So, if your polygon is very small, the automatic `map.fitBounds(bounds)` might respect the desired need of fitting it into the map, but might (for certain users that want the "context" basically render nothing as tile. I know its a fine grained need so will follow up with a different ISSUE that can wait until next release and won't affect by default this one at all

DiegoPino (changes requested):
Small performance review/check

DiegoPino:
@markpbaggett I tagged you a few weeks ago. All looks good but the const v/s var change request would be great for performance needs. Let me know if you have any questions about it. Thanks for your contributions!

DiegoPino:
Ok, did not get a response. Will merge as it is BC if not I will start adding code over it and then we will have to rebase. I will do on top later on the performance optimizations

Issue #590, opened 2026-05-01
Document and normalize data_api.request_date (and all other arguments)


DiegoPino:
# What?

My bad. By default `data_api.request_date` is actually Time `(H:M:i)`. Should be the whole date with Time Zone offset.

That is how I coded it. Bad

Work around is

In your OAI Wrapper template instead of 

```twig
 <responseDate>{{  data_api.request_date }}</responseDate>
```

Use
```twig
{% set response_date = "now"|date("Y-m-d\TH:i:s\Z") %}
{% if data_api.request_date %}
  {% set response_date = "now"|date("Y-m-d") ~ "T" ~ data_api.request_date|render ~ "now"|date("P") %}
{% endif %}
 <responseDate>{{ response_date }}</responseDate>
```

The real solution is to actually return date("Y-m-d\TH:i:s\Z")  from the controller as context which requires for those with custom twig templates manual changes (sorry @alliomeria I hate to do that, but its needed)

Also:

Arguments that contain `:` are wrongly split. I need a different splitting code (view:display:argument:value:api_argument) logic that splits the first 3, then the rest in reverse, joins what is not the last.

Issue #586, opened 2026-04-21
Add config settings for ScrollTop Command triggered on AJAX views (controller) when there is a pager in place


DiegoPino:
# What?

Very fine grained request. Most of our users don't even know about the ADO to ADO AJAX (views <-> html data properties <-> Viewers) nor that we override Drupal's AJAX (which has so many issues) Controller with our own implementation. 

Right now, there is a JS action that triggers an AjAX Command (scroll up) when a view is loaded dynamically && has a pager(even if not used). Our interactions tap in between but eventually call the Drupal AJAX mechanic, so that Command could be triggered.

The idea would be to add an extra setting (like the one we have that allows us to skip setting the URL on an AJAX call) that allows a specific view to not call scroll up at all. Maybe? even a CSS selector option so the scroll-up could also be defined to scroll just to a certain part?

Very niche

DiegoPino:
The specific call is here FYI (to myself, I write/code/do all this just for myself) https://github.com/esmero/format_strawberryfield/blob/12e363482c1a3ae5bf3af71dd122ec2a69ea2918/modules/format_strawberryfield_views/src/Controller/FormatStrawberryfieldViewAjaxController.php#L253-L256

DiegoPino:
See https://github.com/esmero/format_strawberryfield/pull/591/commits/55e67f1c5c981eb7881bd3c01f0946fb596f3cc1
Will be merged with that pull. Once I have the 1.7.0 version and 2.1.0 one merged I can close this one @alliomeria

PR #591, opened 2026-05-04
ISSUE-590: Better Splitting of arguments + full  date in data_api.request_date + SBF driven APIs

DiegoPino:
See #590 

Also:

Anyone that has right now this in an OAI Wrapper template
```Twig
{% set response_date = "now"|date("Y-m-d\TH:i:s\Z") %} {% if data_api.request_date %}
  {% set response_date = "now"|date("Y-m-d") ~ "T" ~ data_api.request_date|render ~ "now"|date("P") %}
{% endif %}
 <responseDate>{{ response_date }}</responseDate>
```

Needs to change it to (once released) to
```
<responseDate>{{  data_api.request_date }}</responseDate>
```

This will go into the release notes

Also (Updated Cinco de Mayo del 2026 🇲🇽 )

Strawberry Flavors can also now feed the API. Means you can build OCR based APIs, annotation based APIs, color based APIs. It will load the parent-to-the-flavor (the one holding the source == ADO) and pass as context to the template (item level one) but also the full Flavor data structure (as Ingested/not only what is in Solr). Allowing one to do extra validation/rendering/etc

Finally. Better cache/access & IP embargo aware response was added and some small Typo + clarification notes on the Actual Config form.


alliomeria:
Makes sense, great fix @DiegoPino!

DiegoPino:
Also committing code for https://github.com/esmero/format_strawberryfield/issues/586. With Symfony 7.4, some Request -> Get methods are also deprecated (Drupal 11.3>) so I will test this first, cherry pick for 1.7.0/then fix that only for 2.1.0 @alliomeria

